PR #29, opened 2020-01-09
Home page

monodo:
Add link to new permit request, start reshaping the homepage


permits/templates/permits/permit_request_new.html
```diff
@@ -1,8 +1,12 @@
 {% extends "base_generic.html" %}
 {% load bootstrap4 %}
 {% load i18n %}
+{% block navbar %}
+{% load permit_progressbar %}
```
sephii: Move the `{% load permit_progressbar %}` out of the navbar block (put it with the other `load` tags).

permits/templates/permits/permit_request_new.html
```diff
@@ -1,8 +1,12 @@
...
+{% endblock navbar %}
 
 {% block content %}
+{% permit_progressbar 1 %}
```
sephii: You'll want to use `{% permit_progressbar permit_request %}` to use the current permit request.

permits/views.py
```diff
@@ -3,7 +3,7 @@
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 
-from . import forms, models
+from . import forms, models, templatetags
```
sephii: Don't import the templatetags in the views. Django loads them with the `{% load permit_progressbar %}` tag.

templates/geomapshark/index.html
```diff
@@ -9,19 +9,42 @@
...
-        {% if user.groups.all.0.name == "extern" %}
 
-          <li><a class="btn btn-primary responsive-btn" href="{% url 'gpf:listexterns' %}"><i class="fa fa-road"></i>&nbsp;Permis de fouille</a></li>
-          <li><a class="btn btn-primary responsive-btn" href="{{apps.mapnv}}" target="_blank"><i class="fa fa-map"></i>&nbsp;mapnv: Géoportail</a></li>
-          <li><a class="btn btn-primary responsive-btn" href="/lidar" target="_blank"><i class="fa fa-cube"></i>&nbsp;mapn: Géoportail 3D (LiDAR)</a></li>
+{% block content %}
+
+
+
+        {% if user.groups.all.0.name == "extern" %}
```
sephii: Out of scope (since this code was already here), but still: this is fragile because if the user has more than 1 group, there's no guarantee the `extern` group will be in first position. You'd better use a custom template tag to render this. You'll then be able to execute some specific code to check if the user is in the `extern` group (eg. `request.user.groups.filter(name='extern').exists()`).

Strictly Django speaking though, you should use [permissions](https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#django.contrib.auth.models.User.has_perm) instead of checking if a user is in a specific group.

geomapshark/static/css/gpf.css
```diff
@@ -12,6 +12,16 @@
   border: 2px solid rgba(252, 200, 70, 1);
 }
 
+#homepage_main_apps {
```
sephii: Why is this duplicated? If this file is copied here by `collectstatic`, it shouldn't be versioned.

monodo:
Don't close it for now, I'll address this comment later @sephii

PR #48, opened 2020-04-14
YGDP-24 vue en lecture seule

sephii:


permits/tables.py
```diff
@@ -6,9 +6,7 @@
 
 
 class PermitRequestTableExterns(tables.Table):
-    edit_properties = tables.TemplateColumn(template_name="tables/_edit_permit_request.html", verbose_name=_('Modifier'), orderable=False)
-    submit = tables.TemplateColumn(template_name="tables/_submit_permit_request.html", verbose_name=_('Envoyer'), orderable=False)
-    delete = tables.TemplateColumn(template_name="tables/_delete_permit_request.html", verbose_name=_('Supprimer'), orderable=False)
+    actions = tables.TemplateColumn(template_name="tables/_permit_request_actions.html", verbose_name=_('Actions'), orderable=False)
```
monodo, line 9: sweet ;-)

permits/exceptions.py
```diff
@@ -0,0 +1,6 @@
+class BadPermitRequestStatus(Exception):
```
monodo, line 1: Cool. Where is the content of the error defined so that it's understandable by the user ?

permits/templates/permits/_permit_request_summary.html
```diff
@@ -0,0 +1,49 @@
...
+          <a href="{{ field.value.url }}" target="_blank">{{ field.value.name|basename }}</a>
+        {% endif %}
+      </dd>
+    {% endfor %}
+  </dl>
+{% endfor %}
+
+{% if contacts %}
+  <h4>{% trans "Contacts" %}</h4>
+  {% for contact_type, contact_infos in contacts %}
```
monodo, line 33: ok thanks

permits/exceptions.py
sephii: The user should never see this exception, since it's caught in `redirect_bad_status_to_readonly` and redirects the user to the readonly view of the permit request. I added a more comprehensible `__str__` anyway, to help developers.

PR #49, opened 2020-04-16
YGDP-9 add permit request submission notification

sephii:


gpf/models.py
```diff
@@ -1,262 +1,262 @@
-# Create your models here.
```
monodo, line 1: Why is this file affected by this PR ?

sephii: I fixed the `related_name` attribute of `Department.administrative_entity`:

`administrative_entity = models.ForeignKey(AdministrativeEntity, null=True, on_delete=models.SET_NULL, related_name='departments', verbose_name=_("administrative_entity"))`

(`related_name` is the name of the accessor for the other side of the relation, so it's used to retrieve the departments in an administrative entity, so the correct name is `departments`)

The weird looking diff comes from the file that contained Windows-style newlines, and when I saved it it changed to Unix-style newlines.

monodo:
ok for me

PR #55, opened 2020-04-27
Normalize line ending and add binary png in .gitattributes.

nicolas-sitylb:
It would be really nice if we can detail all individual file formats that are, indeed text files, in the .gitattribute file.    

A more general way would be to use https://editorconfig.org/ and its plugin EditorConfig with our editors and setup `.gitattributes` along with a `.editorconfig` file at the start of a project.     

The error with unstagged file will always be there if we start back to an older commit. Unless someone want to fix in through the entire git history (...).


monodo:
Merged in YGDP-8

PR #64, opened 2020-05-05
YGDP-11 amender la demande

sephii:
https://jira.liip.ch/browse/YGDP-11


monodo:
missing tablib dependency @sephii

PR #67, opened 2020-05-12
gisserver 3.4 to 3.10 LTS and refactor geo server directory

monodo:


monodo:
@sephii mergin into master. No impact for you

PR #95, opened 2020-06-16
Sprint 6/ygdp 61

nicolas-sitylb:
YGDP-61 finished.


monodo:
Setting phone field max length to 16 brakes the test. Faker uses US phone numbers... @nicolas-sitylb

monodo:
I'll set this back to 19

monodo:
Also, please never edit directly inital migration...

PR #106, opened 2020-08-11
Ygdp 83

nicolas-sitylb:
Pre-merge request for printpermit.
There is still some things to be fixed and debugged.


nicolas-sitylb:
todo:
- enhance multiple requests
- add "service pilote" values ; where are they?
- set CSS

monodo:
@nicolas-sitylb FILTER is now ok, but BBOX extent has to be fixed

PR #111, opened 2020-09-24
Configure availables status per administrative entities and hide unused workflow steps

monodo:
Configure available status for givent administrative entity from admin


monodo:
@rbovard
Still to do:
- Filter the tabs depending on status
- Improve UI in admin as it's quite confusing right now

monodo:
The application logic is changedn in views.get_request_validation_form, views.get_validation_form, views.get_poke_form. This breaks many test and we'll need to discuss the best way to rewrite them with @sephii

monodo:
We'll open new PR right way

PR #112, opened 2020-09-25
Ygdp 25

monodo:
Log changes in tables


monodo:
replaced by django-simple-history

PR #116, opened 2020-09-29
YGDP-86: add a button for the secretariat to be able to modify a permit request.

nicolas-sitylb:


nicolas-sitylb:
TODOs in code probably want to be completed:    
https://github.com/yverdon/geocity/blob/master/permits/views.py#L246    
https://github.com/yverdon/geocity/blob/master/permits/views.py#L317

PR #117, opened 2020-10-06
Kml import

monodo:


monodo:
@nicolas-sitylb Polygon import is missing

PR #119, opened 2020-10-08
Ygdp 93 - KML import in user and admin interface

monodo:


monodo:
@sephii no impact on application logic, pure geo functionnalities, thus merging so that you start YC-4 from an up to date version

PR #122, opened 2020-10-20
Add sub urls support

monodo:
Add support for sub url config like mydomain/my_alias
Cleanup docker-compose setup


env.demo
```diff
@@ -1,5 +1,13 @@
 
-DEBUG=True
+DEBUG=True # SET TO False in PRODUCTION!
+DEV_DEPENDENCIES="true" # MUST BE COHERENT WITH DEBUG value (DEBUG=True => DEV_DEPENDENCIES="true")
+DJANGO_DOCKER_PORT=9095 # The django application container
+QGISSERVER_DOCKER_PORT=9096 # The qgisserver osgeo container
+POSTGRES_DOCKER_PORT=9097 # The postgres container DEV ONLY
+STATIC_URL = '/static/'
+PREFIX_URL = 'geocity_test' # Set your lovely url prefix here (mydomain.ch/prefix_url/)
```
rbovard, line 8:
```suggestion
PREFIX_URL = 'geocity_test/' # Set your url prefix here (mydomain.ch/<prefix_url/>)
```

env.demo
```diff
@@ -1,5 +1,13 @@
 
-DEBUG=True
+DEBUG=True # SET TO False in PRODUCTION!
+DEV_DEPENDENCIES="true" # MUST BE COHERENT WITH DEBUG value (DEBUG=True => DEV_DEPENDENCIES="true")
+DJANGO_DOCKER_PORT=9095 # The django application container
+QGISSERVER_DOCKER_PORT=9096 # The qgisserver osgeo container
+POSTGRES_DOCKER_PORT=9097 # The postgres container DEV ONLY
+STATIC_URL = '/static/'
```
rbovard, line 7:
```suggestion
STATIC_URL = '${PREFIX_URL}static/'
```

rbovard:
I suggested two changes in url variables to make it working.

There are still some issues with assets in `accounts/login/static`, you can show them in my test instance.

env.demo
rbovard: If we keep the use of `${PREFIX_URL}` here, we should invert line 7 and 8.

monodo: Well, ´´´${PREFIX_URL}´´´ is just an indication at this level as it will not be interpreted at this level. ´´´STATIC_URL´´´ can could be change in many different manners as some integrator might want to move all static assets outside the project and not serve theme using the default gunicorn + whitenoise solution. ´´´collectstatic´´´ django managment command can be use for that purpose

monodo:
@nicolas-sitylb please review, test on srvdev & merge. Much easier docker setup from now on

PR #123, opened 2020-10-20
Fix some stuff

rbovard:
- Fix rights for entrypoint_dev.sh (same rights as for entrypoint.sh)
- Fix indent in docker-compose.yml
- Update doc


monodo:
LGTM

PR #132, opened 2020-11-09
Add WORKObjects-related WMS layer to map

monodo:


monodo:
@nicolas-sitylb please review

static/customWidgets/GeometryWidget/geometrywidget.js
```diff
@@ -35,6 +34,35 @@
...
+    ;
+
+    var layers =  this.options.wms_layers.split(',');
+    var ol_layers = [];
+    for (var i=0; i < layers.length; i++) {
+        ol_layers.push(
+            new ol.layer.Image({
+            source: new ol.source.ImageWMS({
+              url: layers[i],
+              projection: "EPSG:2056",
```
nicolas-sitylb: What happens if a layer in the groups comes with an other projection system? Maybe it's more robust to parse and extract the CRS parameter from the WMS URL?

static/customWidgets/GeometryWidget/geometrywidget.js
```diff
@@ -34,33 +34,22 @@
...
-            opacity: 1,
-          })
+    var layer_list = [];
+
+    this.options.wms_layers.split(",").forEach(function (item, index) {
+      layer_list.push(
+        new ol.layer.Image({
+          source: new ol.source.ImageWMS({
+            url: item,
+            projection: "EPSG:2056",
```
nicolas-sitylb, line 44: Use CRS parameter from WMS here?

nicolas-sitylb (approved):

static/customWidgets/GeometryWidget/geometrywidget.js
monodo: => The whole widget is for EPSG:2056

nicolas-sitylb: Then what if people want to upload WGS84 KML files? Or WorldMercator GPX data? Maybe we should consider that in the future as well?

monodo: => KML are reprojected on import => https://github.com/yverdon/geocity/blob/master/static/customWidgets/GeometryWidget/geometrywidget.js#L308

Issue #139, opened 2020-11-23
Unable to start instance in dev mode


nicolas-sitylb:
Docker is always installing dependencies in "production" mode, hence not using the python dependencies specified in `requirements_dev.txt` file.

PR #140, opened 2020-11-23
Fix dockerfile and set a global ENV var for development purpose

nicolas-sitylb:
Fixes https://github.com/yverdon/geocity/issues/139


monodo:
This PR breaks the build.

nicolas-sitylb:
This fix is intended to fix the build in `DEV` mode, and it does.    
So what exactly is broken?

nicolas-sitylb:
@monodo it's OK; errors on the `master `are also fixed in this branch now.

PR #141, opened 2020-11-24
YC-4 support for multiple permit geo times

sephii:


monodo:
@sephii 
Look very good :-) . 2 issues:

1. Load KML data outputs following error "Uncaught TypeError: this.el.querySelector(...) is null" looks like `<input type="file" class="custom-file-input" id="selectfile" accept=".kml" data-action="addKML">` is causing it.

2. Permit summary: will need some adaptations in order to use the new version of the widget.

permits/templates/geometrywidget/geometrywidget.html
```diff
@@ -1,122 +1,93 @@
...
-{% block stylesheets%}
-  {% load fontawesome %}
-  {% fontawesome_stylesheet %}
-    <link rel="stylesheet" type="text/css" href="{% static 'css/geotime.css' %}">
-{% endblock %}
-
+{% load permits_extras %}
 
 <style type="text/css">{% block map_css %}{% get_current_language_bidi as LANGUAGE_BIDI %}
+  {# TODO cleanup #}
```
monodo, line 8: TODO ?

permits/views.py
```diff
@@ -496,24 +498,34 @@ def permit_request_actors(request, permit_request_id):
...
 @login_required
 def permit_request_geo_time(request, permit_request_id):
-
     permit_request = services.get_permit_request_for_user_or_404(request.user, permit_request_id)
+    PermitRequestGeoTimeFormSet = modelformset_factory(
+        models.PermitRequestGeoTime,
+        form=forms.PermitRequestGeoTimeForm,
+        extra=0,
+        min_num=1,
+        can_delete=True
```
monodo, line 507: Appears as a checkbox in form. Any simple mean to make it look like something more obvious for the user ?

monodo (commented):
Looks good except for 2 little details

permits/templates/geometrywidget/geometrywidget.html
sephii: Indeed, I still need to clean this up.

permits/views.py
sephii: No simple way I’m afraid, that’s the way Django does it by default with fieldsets. We could replace this by something else, but it’s not that simple.

monodo: Ok, no worries

static/js/permits_geo_time.js
```diff
@@ -0,0 +1,126 @@
...
+      this.addButtonNode.addEventListener("click", this.addForm.bind(this));
+
+      [...this.formsContainerNode.querySelectorAll(".collapse")].forEach(
+        this._fixMapSizeOnCollapseOpen
+      );
+    }
+
+    _fixMapSizeOnCollapseOpen(collapseNode) {
+      jQuery(collapseNode).on("shown.bs.collapse", (e) => {
+        let geometryWidgetNode = collapseNode.querySelector("[data-geometry-widget]");
```
ludovicturmel, line 81: Throw an error in the console here. Selector return null

ludovicturmel: Need to check once the issu is fixed if we can add manually coordinate because for now it's seem to be broken

permits/templates/permits/_geo_time_form.html
```diff
@@ -0,0 +1,15 @@
+{% load i18n %}
+{% load bootstrap4 %}
+
+<div class="card mb-3">
+  <div class="card-body">
+    <a data-toggle="collapse" href="#planning-{{ form_id }}">
```
ludovicturmel, line 6: For this collapse who be easier to understand if we add arrow down & up. So it's easier to understand for the end-user the collapse action. 

On the same topic it's not really nice if a collapse is close when the user try to go to the next step but there is some error inside the form. The errors on the fields are hidden inside the collapse content.

sephii: I agree it would be nice to have arrows here, but we don’t have icons on the projet yet and it’s not in the acceptance criteria, so I’d say it’s good enough for now, and we’ll add some once we add icons to the project.

The collapse is now open if the form has errors.

PR #147, opened 2020-12-01
use constance for steps titles

monodo:


loicgasser:
Hey, instead of adding additional settings for the title of the steps. Wouldn't it be better to simply add these in the locale files. Then when you need to create a new application, you only need to edit the locale files.

PR #149, opened 2020-12-07
[YGDP-76] signup signin process & 2FA 

loicgasser:
What has been done:
- Fix env.demo for variable substitution in docker-compose
- Install and integrate [django-two-factor-auth](https://github.com/Bouke/django-two-factor-auth)
- Fix multiple CSS issues on mobile
- Multiple other CSS/HTML improvements

What needs to be done:
- [x] Untangle the mess of the admin site see: https://github.com/Bouke/django-two-factor-auth/issues/219
- [x] Make 2FA optional thanks to an environment variable.
- [x] Create tests for both type of applications.


monodo:
merged in main, thanks @loicgasser !!

PR #151, opened 2020-12-11
GitHub CI

loicgasser:
This PR has been built on top of https://github.com/yverdon/geocity/pull/149.

PR #152, opened 2020-12-11
Django upgrade to version 3

loicgasser:
It upgrades Django to the latest major stable version. We also upgrade GDAL from version 3.1 to 3.2, as a result I also updated the docker image. I push it to my account for now, but we'll push it to Yverdon account.

This PR has been built on top of https://github.com/yverdon/geocity/pull/151


monodo:
thanks @loicgasser merged in main :-)

PR #153, opened 2020-12-16
YC-78 change cell color depending on status

anthonyledda:


monodo:
Thanks @anthonyledda merged in main

PR #243, opened 2021-06-02
hide geocaledar specific fields and buttons

monodo:


geomapshark/settings.py
```diff
@@ -166,6 +167,11 @@
         "URL de l'application calendrier cartographique",
         str,
     ),
+    "ENABLE_GEOCALENDAR": (
+        True,
+        "Définir si l'application du calendrier cartographique est utilisée ou pas (doit dans tous les cas être installé à part)",
```
rbovard, line 172:
```suggestion
        "Définir si l'application du calendrier cartographique est utilisée ou pas (doit dans tous les cas être installée à part)",
```

rbovard (changes requested):
Not tested live, but I think you forgot the publish field in "Traiter" tab

PR #244, opened 2021-06-02
Yc 240

monodo:


rbovard:
In the step validation approval `/approve` I still see the publish checkbox:

image

@monodo Are we talking of the same publish field?

monodo:
Sounds quite bad

monodo:
Ok, that's juste one more template to adapt.

PR #260, opened 2021-06-16
add type to contact type in order to avoid wrong autofill by chrome

monodo:


AlexandreJunod:
Not working

rbovard:
It's better.

On the second contact it's working on field "Prénom" but not "Nom".

But after for next contacts, it proposes me CC infos for both "Prénom" and "Nom".

PR #261, opened 2021-06-17
Token helper for oauth2 application tokens

AlexandreJunod:
Used for qgis templates


monodo:
WFS 3 implementation that will be donw shortly makes this helper no longer useful

PR #268, opened 2021-06-21
Fix administrative entity

monodo:


monodo:
Test runned locally OK, thus merging

PR #269, opened 2021-06-21
add validations et refactor actors

monodo:


monodo:
tests ok , thus merging

PR #283, opened 2021-07-06
YC-357 Add mandatory 2FA for groups users

felicemcc:


permits/views.py
```diff
@@ -890,6 +891,14 @@ class PermitRequestList(SingleTableMixin, FilterView):
     paginate_by = int(os.environ["PAGINATE_BY"])
     template_name = "permits/permit_requests_list.html"
 
+    def render_to_response(self, context):
+        if (
+            services.is_2FA_mandatory(self.request.user)
+            and not self.request.user.is_verified()
+        ):
```
T4m, line 898: I think this a wrong place for this check. It only enforces 2FA on this specific view. I think this condition&redirection should be done in a middleware or called on login (regardless the login's "next" view). Still looking at it.

T4m: I think the best approach would be to extend the decorator: **otp_required**, so it checks the group configuration, and when needed calls its super() method. Then add right after every login_required decorators. WDYT?

permits/tests/test_c_admin_site.py
```diff
@@ -249,3 +286,62 @@ def test_cannot_add_a_new_integrator_group_to_an_user_who_has_already_an_integra
...
+    if settings.ENABLE_2FA:
+
+        def test_user_of_group_with_mandatory_2FA_is_asked_to_set_it_up(self):
+            user = SecretariatUserFactory()
+            user.groups.set([self.group2fa])
+            self.client.login(username=user.username, password="password")
+
+            response = self.client.get(
+                reverse(settings.LOGIN_REDIRECT_URL), follow=True,
+            )
```
T4m, line 299: I think it's good, but the user might choose another redirect with the "next" argument or by entering any path in the URL bar after the redirection. In this case, shouldn't he be enforced to use 2FA as well? (Related to a comment below)

permits/views.py
felicemcc: Moved the check to a decorator used after every login_required, works like a charm...

permits/tests/test_c_admin_site.py
felicemcc: Solved by the decorator check

permits/views.py
```diff
@@ -117,7 +118,26 @@ def progress_bar_context(request, permit_request, current_step_type):
     return {"steps": steps, "previous_step": previous_step}
 
 
+def check_mandatory_2FA(
+    view=None, redirect_field_name="next", login_url="profile", if_configured=False
```
T4m, line 122: `is_configured` au lieu de `if_configured` non ?

felicemcc: It is ok, adopted from the original **otp_required** method, it means If ``True``, an authenticated user with no confirmed
 OTP devices will be allowed. Default is ``False``

PR #285, opened 2021-07-07
Sprint-7/yc-337 Admin can select if a validation document is required

felicemcc:


permits/tests/test_a_permit_request.py
```diff
@@ -2014,6 +2014,85 @@ def test_classify_sets_validation_date(self):
...
+            ),
+        )
+
+        self.assertContains(response, "Approbation de la demande")
+        self.assertEqual(
+            validation.permit_request.status,
+            models.PermitRequest.STATUS_AWAITING_VALIDATION,
+        )
+        self.assertContains(response, "validation_pdf")
+
```
T4m, line 2095: I would a specific TestCase class for two reasons:
- Remove useless actions on setUp (increase tests speed)
- Add redundant actions in setUp (less lines in test methods)

I'm doing a proposal.

PR #288, opened 2021-07-13
Sprint-7/YC-362 Send email to concerned services on Secretariat Approval/Rejection of permit request

felicemcc:


permits/views.py
```diff
@@ -1171,6 +1172,33 @@ def permit_request_classify(request, permit_request_id, approve):
...
+            works_object_types_to_notify = permit_request.works_object_types.filter(
+                notify_services=True
+            )
+
+            if works_object_types_to_notify.exists():
+                mailing_list = []
+                for emails in works_object_types_to_notify.values_list(
+                    "services_to_notify", flat=True
+                ):
+                    emails_addresses = emails.replace(" ", "").split(",")
```
T4m, line 1185: There's a `strip()` function on strings which does that replacement on extremities. Is that what you want?

We shouldn't replace spaces inside the email, which might be valid in very rare cases (like `foo\ bar@space.com`).

permits/migrations/0032_add_notification_for_services_in_wot.py
```diff
@@ -0,0 +1,23 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='worksobjecttype',
+            name='notify_services',
+            field=models.BooleanField(default=False, verbose_name='Notifier les services'),
+        ),
+        migrations.AddField(
+            model_name='worksobjecttype',
+            name='services_to_notify',
+            field=models.TextField(blank=True, verbose_name='Emails des services à notifier'),
```
T4m, line 21: It seems that if you use space or line return, the emails won't be separated correctly in the permits view.
So I would add a helptext : 
Veuillez séparer les emails par une virgule ","

permits/views.py
```diff
@@ -1171,6 +1172,33 @@ def permit_request_classify(request, permit_request_id, approve):
...
+                    mailing_list += [
+                        ea for ea in emails_addresses if services.validate_email(ea)
```
T4m, line 1187: => `ea.strip() for ea in ...`

So, we remove spaces on extremities only.

WDYT?

permits/views.py
felicemcc: Ok, better to use strip() after....
I added the help text, but just to avoid human error for newlines, we are covered by changing to this line to: 
`emails_addresses = emails.replace("\n", ",").split(",")`

permits/migrations/0032_add_notification_for_services_in_wot.py
felicemcc: Done!

PR #325, opened 2021-09-28
[DRAFT] OGC Features API (wfs3) endpoint

olivierdalang:
This is the WIP implementation for WFS3 (aka OGC Features API) for Django applied to Geocity.

Please see `django_wfs3/README.md` for more info on implementation.

geocitywfs3

It's not ready yet, but opening this PR already so you can review and see if it takes the desired direction.

Roadmap :
- [x] basic implementation
- [x] working readonly endpoints 
  - [x] PermitRequestViewSet (incl. subclasses for point/line/poly only geometries so they can load in QGIS)
  - [x] PermitRequestGeoTimeViewSet
- [ ] fix CRS in QGIS (seems CRS is simply ignored - do we have a working example, or is it an issue with QGIS ?)
- [x] various other fixes (see TODO in code)
- [x] test with oauth2 (it should work already, but unsure how this must be configured from QGIS)

### DIFF
--- a/django_wfs3/README.md
+++ b/django_wfs3/README.md
@@ -54,5 +54,4 @@ WFS3_DESCRIPTION = "Description"
 
 This is probably still relatively far from a full OGC Services API implementation and currently only aims to support read-only view from QGIS.
 
-This app will as some point be factored out
-into a reusable Django library.
+This app will at some point be factored out into a reusable Django library.
--- a/django_wfs3/README.md
+++ b/django_wfs3/README.md
@@ -8,22 +8,46 @@ serialization, authentication, etc.).
 
 ## Usage
 
+> NOTE : these snippets are not tested and may require fixing/adaptations.
+
 Add this to your `urls.py` :
 
 Register your viewset in the wfs3 router
 ```python
-
+from rest_framework_gis.serializers import GeoFeatureModelSerializer
 from django_wfs3.urls import wfs3_router
-from .viewsets import MyModelViewSet
+
+class MyModelSerializer(gis_serializers.GeoFeatureModelSerializer):
+    class Meta:
+        model = MyModel
+        fields = "__all__"
+        geo_field = "geom"
+
+class MyModelViewset(WFS3DescribeModelViewSetMixin, viewsets.ModelViewSet):
+    queryset = MyModel.objects.all()
+    serializer_class = MyModelSerializer
+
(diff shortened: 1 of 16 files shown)


monodo:
If environment variable PREFIX_URL has a value, it is missing in the first link:

image

olivierdalang:
> If environment variable PREFIX_URL has a value, it is missing in the first link:

Good catch, should be fixed.

olivierdalang:
Made the changes as discussed :
- removed geotime endpoints
- added permits endpoints with specific geometries extracts

image

monodo:
please go into web container and execute `black .`

olivierdalang:
Added a new commit that makes layer homogeneous, which hopefully will fix loading of prod data in QGIS. This changes a bit the behavior, so please review carefully the following changes :

- We now return null instead of an empty point when there's no geometry. Unsure why it used to be an empty point  ? Depending on how/where this is used, it could break existing behavior.

- We now add a small padding to bounding boxes, to ensure it returns a polygon (BOX2Ds representation are not necessarily polygons). This changes the representation, IMO it's even more clear with the small padding. Currently set to 1m, we could also make that very low if we don't want a visible padding.

image

django_wfs3/README.md
```diff
@@ -0,0 +1,58 @@
...
+```python
+WFS3_TITLE = "My Endpoint"
+WFS3_DESCRIPTION = "Description"
+```
+
+## Roadmap / status
+
+This is probably still relatively far from a full OGC Services API implementation and currently only aims to support read-only view from QGIS.
+
+This app will as some point be factored out
```
monodo, line 57: typo "at"

geomapshark/settings.py
```diff
@@ -370,5 +370,11 @@
     "DEFAULT_AUTHENTICATION_CLASSES": (
         "oauth2_provider.contrib.rest_framework.OAuth2Authentication",
         "rest_framework.authentication.SessionAuthentication",
-    )
+        # TODO : remove this once OAuth2 settings from qgis are clear
+        "rest_framework.authentication.BasicAuthentication",
```
monodo, line 374: oAuth2 with QGIS has been tested and workd fine, thus, you can remove this line @olivierdalang

monodo (commented):
I'll push this little directly in your branch and merge right ahead

PR #330, opened 2021-10-13
YC-520  send email new permit notification to user that are both backoffice and validator

monodo:
Also remove notification to integrator_admin


rbovard (approved):
All good with my usecases:
- Secretariat and validator distinct users
- Secretariat and validator unique user

Thanks a lot for the quick fix.

PR #365, opened 2021-11-17
user must have permission to classify in order to see the print tab

monodo:


permits/templates/permits/_permit_request_actions.html
```diff
@@ -54,12 +54,13 @@ <h2>{% trans "" %}</h2>
 
       {% endif %}
 
-
-          <li class="nav-item">
-          <a class="btn btn-primary btn-action nav-link" id="print-tab" data-toggle="tab" href="#print" role="tab" aria-controls="print">
-              {% trans "Imprimer" %}
-          </a>
-          </li>
+      {% if has_permission_to_classify %}
```
rbovard, line 57: Should't be `has_permission_to_classify_permit_request`?

monodo: It's just the short name used to pass the variable to the context:

`has_permission_to_classify": services.has_permission_to_classify_permit_request`

PR #372, opened 2021-12-09
filter out users if domains or exceptions are not configured by admin

monodo:


nicolas-sitylb (changes requested):
Seems good, but I would block domain names which do not correspond to official administrative entity in the country. Put another way, I would only allow <commune/or-other-admin-entity>.ch domain names.

rbovard:
@nicolas-sitylb Have you think about your case yverdon-les-bains.ch & ylb.ch

@monodo Not tested (I'm not able to do it right now), but a suggestion about the texts:
- Liste de domaines séparés par des virgules ',' correspondant aux utilisateurs rattachés à l'entité administrative (ex: yverdon-les-bains.ch,ylb.ch)
- Liste d'emails séparés par des virgules ',' d'utilisateurs spécifiques rattachés à l'entité administrative (ex: greffe@nowhere.com)

PR #373, opened 2021-12-13
Span repr (title) element over the whole table

nicolas-sitylb:
Span repr (title) element over the whole table and add virtual dd element to keep the rest of the table aligned.


nicolas-sitylb:
image

PR #378, opened 2021-12-16
Improve map widget

rbovard:
First shot, what do you think?

Before
image

After
image


monodo:
Looks good, thanks. I've checked, it'a all ok in both user form and admin 👍

rbovard:
Great, if nothing else you can merge :)

PR #382, opened 2022-01-10
Yc 342

nicolas-sitylb:
The User can still receive e-mails when he publishes a request, I cannot figure out why.

-> Found the [culprit](https://github.com/yverdon/geocity/blob/f4844ff7daaf89e9a414fed4a24ba0c2486d42ae/permits/services.py#L983):

```
data["users_to_notify"] = [permit_request.author.user.email]
```

So it SHOULD be fine with that.

### DIFF
--- a/permits/tests/test_a_permit_request.py
+++ b/permits/tests/test_a_permit_request.py
@@ -2826,10 +2826,6 @@ class PermitRequestValidationRequestTestcase(LoggedInSecretariatMixin, TestCase)
         )
 
         permit_request.refresh_from_db()
-        # test mails
-
-        # self.assertEqual(len(mail.outbox), 2)
-        # self.assertEqual(mail.outbox[0].to, [validator_user.permitauthor.user.email])
 
         self.assertEqual(
             permit_request.status, models.PermitRequest.STATUS_AWAITING_VALIDATION


monodo:
You need to filter emails in service.py. Looks that servicep.py modifications where not commited

permits/services.py
```diff
@@ -1001,14 +1004,25 @@ def request_permit_request_validation(permit_request, departments, absolute_uri_
         models.PermitRequestValidation.objects.get_or_create(
             permit_request=permit_request, department=department
         )
-
-    users_to_notify = {
-        email
-        for department in departments
-        for email in department.group.user_set.values_list(
-            "permitauthor__user__email", flat=True
+    if False:
```
nicolas-sitylb, line 1007: For the moment, I've set up that switch here because I'm not quite sure that the new version does exactly the same, especially that the variable `groups__permitdepartment__in` is equivalent to the for loop with `department.group.user_set` in the previous form.

nicolas-sitylb: Found the [culprit](https://github.com/yverdon/geocity/blob/f4844ff7daaf89e9a414fed4a24ba0c2486d42ae/permits/services.py#L983):

```
data["users_to_notify"] = [permit_request.author.user.email]
```

So it SHOULD be fine with that.

nicolas-sitylb:
@monodo please review and merge if needed.

permits/services.py
nicolas-sitylb: Removed old code in commit [5336e2b3a1542aacebba243d5e8cf78124a7e2a7](https://github.com/yverdon/geocity/pull/382/commits/5336e2b3a1542aacebba243d5e8cf78124a7e2a7).

monodo (changes requested):

README.md
```diff
@@ -267,6 +267,16 @@ docker-compose exec web python3 manage.py makemigrations --help
 docker-compose exec web python3 manage.py migrate --help
 ```
 
+### Practical example
+Note: prepend `docker-compose exec web ` to the following python calls if your app is containerized.
+
```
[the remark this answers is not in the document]
nicolas-sitylb: As this is project specific "rules", and because they were useful for me, I would keep these lines here.    
I'm also thinking of newcomers to the project, e.g. external academic partners...

monodo (approved):

PR #383, opened 2022-01-11
Yc 384

nicolas-sitylb:
Fix # 384 in Jira.


permits/admin.py
```diff
@@ -767,6 +840,14 @@ def __new__(cls, *args, **kwargs):
 
 
 class PermitAdministrativeEntityAdminForm(forms.ModelForm):
+    """ Form class to configure an administrative entity (commune, organisation)
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(PermitAdministrativeEntityAdminForm, self).__init__(*args, **kwargs)
+        self.fields["expeditor_name"].disabled = False
+        self.fields["expeditor_name"].initial = None
```
monodo, line 849: useful ? if not, please remove

monodo (changes requested):

permits/admin.py
nicolas-sitylb: Done.

permits/admin.py
```diff
@@ -223,6 +228,17 @@ def get_queryset(self, request):
...
         return qs
 
+    def save_model(self, req, obj, form, change):
+        """ Set 'is_staff=True' when the saved user is in a integrator group.
+        """
+        if req.user.is_superuser:
+            obj.is_staff = False
+            for group in form.cleaned_data["groups"]:
+                if group.permitdepartment.is_integrator_admin:
+                    obj.is_staff = True
```
monodo, line 238: If `obj.is_superadmin`, do nothing. Otherwise, superadmin could loose it's admin access

permits/admin.py
```diff
@@ -364,11 +380,45 @@ def clean_permissions(self):
...
     inlines = (PermitDepartmentInline,)
     form = GroupAdminForm
+    list_display = [
+        "__str__",
+        "get__is_validator",
+        "get__is_default_validator",
+        "get__integrator",
+        "get__mandatory_2fa",
+    ]
+
```
monodo, line 390: If not too complicated, change the display to checkbox like in WOTs for instance (1.4)

image

monodo (changes requested):
2 little points to check/fix

permits/admin.py
monodo: https://docs.djangoproject.com/en/4.0/ref/contrib/auth/#django.contrib.auth.models.User.is_superuser

nicolas-sitylb: Resolved by https://github.com/yverdon/geocity/pull/383/commits/7a2f14fd1dac30ea60bd329856740907910bfe15

nicolas-sitylb:
Fixed `is_staff` for super users + synchronized with the `main` branch.    
(also added an `order` field in the WorksObject model which could serve as a basis to order the works objects).

monodo (approved):

PR #384, opened 2022-01-12
filter entities by subdomain name

monodo:


monodo:
replaced by https://github.com/yverdon/geocity/pull/449

PR #386, opened 2022-01-31
Migrate to PG13.

nicolas-sitylb:
Migrates to PG13 in both the development and the demo Compose files.

You will probably have to `down -v` if you have a previous version of a PG database running in order to set it up with the newest version.

### DIFF
--- a/docker-compose-demo.yml
+++ b/docker-compose-demo.yml
@@ -45,7 +45,7 @@ services:
             - PREFIX_URL
 
     postgres:
-        image: postgis/postgis:14-3.2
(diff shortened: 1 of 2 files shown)


monodo (changes requested):
This upgrades to PG14 not PG13 ;-)

nicolas-sitylb:
> This upgrades to PG14 not PG13 ;-)

Indeed. Fixed.

rbovard:
> You will probably have to `down -v` if you have a previous version of a PG database running in order to set it up with the newest version.

You ~will probably~ have to `down -v` ...

PR #390, opened 2022-02-01
YC-563: Display prolongation tab only if request can be prolonged

rbovard:
Fix https://jira.liip.ch/browse/YC-563

Please review and test.

For a reason I don't get, there is no change detected after editing the helper_text:

```
python manage.py makemigrations -n fix_typo
No changes detected
```

### DIFF
--- a/permits/templates/permits/_permit_request_actions.html
+++ b/permits/templates/permits/_permit_request_actions.html
@@ -64,7 +64,7 @@
 
       {% endif %}
 
-      {% if forms.prolong and permit_request.can_be_prolonged %}
+      {% if forms.prolong and prolongation_enabled %}
 
           <li class="nav-item">
               <a class="btn btn-primary btn-action nav-link{% if active_form == 'prolong' %} active{% endif %}" id="classify-tab" data-toggle="tab" href="#prolong" role="tab" aria-controls="prolong" aria-selected="{% if active_form == 'prolong' %}true{% else %}false{% endif %}">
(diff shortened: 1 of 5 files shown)


rbovard:
@monodo Hum, I'll need some help on the failing test...

permits/templates/permits/_permit_request_actions.html
```diff
@@ -156,7 +156,7 @@ <h4 class="my-5">{{ object_type }}</h4>
             </div>
         {% endif %}
 
-        {% if forms.prolong %}
+        {% if forms.prolong and permit_request.can_be_prolonged %}
```
monodo, line 159: `can_be_prolonged` is not the right method to use here. This method checks that it is possible / allowed to ask for prolongation not that there is a prolongation setup for the given WOT

`self.status in self.PROLONGABLE_STATUSES and self.max_validity is not None`

That's one reason that make the test` test_secretariat_prolonge_form_is_disabled_on_bad_permit_request_status` fails: the form should be disable but it's no even there.

The simplest way to check if a WOT has a prolongation setup is to check if  WOTs field `permit_duration` is not NULL (not tested on my side)

This way, test should pass  and I'll recommend to add one more test to ensure that permits with NULL value in permit_duration WOTs field are indeed not visible in backoffice. The reverse test (aka check if a missing form fails the test) seems to work fine, obviously ;-)

monodo (changes requested):

permits/templates/permits/_permit_request_actions.html
rbovard: > The simplest way to check if a WOT has a prolongation setup is to check if WOTs field permit_duration is not NULL (not tested on my side)

I see, thanks. But how do I acces to the the WOT here?

monodo: I would add a variable to the context 'prolongation_enabled' for instance. You'll need to add this variable to the context here: https://github.com/yverdon/geocity/blob/main/permits/services.py#L1277

To get the selected WOT for the permit you can simply call the `get_works_object_type_choices`  helping function

rbovard:
Are the test failure related to your work on the API?

rbovard:
@monodo Tests are fine... I'll open another PR with the fix.

Could you please review?

rbovard:
Tests added, please review

monodo (approved):

PR #392, opened 2022-02-01
YC-589: Improve backoffice tabs

rbovard:
Fix https://jira.liip.ch/browse/YC-589

Please review


rbovard:
@monodo Hum, I'll need some help on the failing test...

permits/templates/permits/_permit_request_actions.html
```diff
@@ -127,10 +131,49 @@ <h4 class="my-5">{{ object_type }}</h4>
...
+                    <button class="btn btn-primary mr-3" formaction="{% url 'permits:permit_request_print' permit_request_id=permit_request.pk template_id=print_template.pk %}">
+                    {{ print_template.description  }}
+                    </button>
+                {% endfor %}
+                </form>
+            {% else %}
+                <span class="no_print_template">{% trans "Aucun modèle d'impression n'a été configuré"%}</span>
+            {% endif %}
+            </div>
+            <h5>{% trans "Directives liées à la demande" %}</h5>
```
monodo, line 151: I wonder: should we not move this in a box better that in a tab ? In the future, the directives will most probably be displayed in a new **documents**  (yc-536) tab. 

A solution would be:
- move the directive in a box
- write test to check the box is visible if directive is set
- remove this part assertion in test https://github.com/yverdon/geocity/blob/acdda61ef3e0fc62d9f4472f05e9f01894cb08b2/permits/tests/test_a_permit_request.py#L2649-L2651
- Adapt the test https://github.com/yverdon/geocity/blob/acdda61ef3e0fc62d9f4472f05e9f01894cb08b2/permits/tests/test_a_permit_request.py#L2622 so that we check that the print tab is not visible if not print template is configured
- Write new test to ensure print tab is visible is print template is confifure

monodo (changes requested):

permits/templates/permits/_permit_request_actions.html
```diff
@@ -34,11 +34,31 @@ <h2>{% trans "" %}</h2>
...
+
+        <li class="nav-item">
+        <a class="btn btn-primary btn-action nav-link" id="print-tab" data-toggle="tab" href="#print" role="tab" aria-controls="print">
+            {% trans "Impression" %}
+        </a>
+        </li>
+
+      {% endif %}
+
+      {% if has_permission_to_classify and directives %}
```
monodo, line 47: Ok, I was thinking about moving the directives into a block like this: https://github.com/yverdon/geocity/blob/main/permits/templates/permits/permit_request_detail.html#L197-L202 but maybe a tab is also fine ?

monodo (approved):

PR #393, opened 2022-02-02
YC-592: Display validation status as radioselect

rbovard:
Fix https://jira.liip.ch/browse/YC-592

Please review.

### DIFF
--- a/permits/forms.py
+++ b/permits/forms.py
@@ -1198,6 +1198,7 @@ class PermitRequestValidationForm(forms.ModelForm):
             "comment_after",
         ]
         widgets = {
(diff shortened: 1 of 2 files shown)


permits/forms.py
```diff
@@ -1208,13 +1209,7 @@ def __init__(self, *args, **kwargs):
 
         # Show "----" instead of "en attente" for the default status
```
monodo, line 1210: remove deprecated comment

monodo (changes requested):

permits/forms.py
rbovard: Good catch

monodo (approved):

PR #400, opened 2022-02-04
Fix never displaying directive addtionnal info

rbovard:
Fix something that was never working... (when only `additional_information` is set)

I used the same logical as in https://github.com/yverdon/geocity/blob/main/permits/templates/permits/permit_request_submit.html#L28-L40


rbovard:
All green, merging.

PR #401, opened 2022-02-04
Yc 591 b

monodo:


permits/models.py
```diff
@@ -431,6 +431,14 @@ class PermitRequest(models.Model):
     status = models.PositiveSmallIntegerField(
         _("état"), choices=STATUS_CHOICES, default=STATUS_DRAFT
     )
+    shortname = models.CharField(
+        _("nom court de la demande, de l'événement,..."),
```
rbovard, line 435:
```suggestion
        _("nom court de la demande, de l'événement, etc."),
```

rbovard: Also adapt the migration

permits/views.py
```diff
@@ -303,10 +303,18 @@ def get_amend_form(self, data=None):
         if services.has_permission_to_amend_permit_request(
             self.request.user, self.permit_request
         ):
+            # Get the first object type selected as a shorname suggestion for pilot
+
```
rbovard: Remove this extraline (and and one before next comment)

permits/api.py
```diff
@@ -76,6 +85,9 @@ def get_queryset(self):
         if starts_at:
             start = datetime.datetime.strptime(starts_at, "%Y-%m-%d")
             base_filter &= Q(starts_at__gte=start)
+        if show_only_future:
+            start = datetime.datetime.now() - datetime.timedelta(days=7)
```
rbovard, line 89: Why not simply filter on events with end date >= today? If yes, also adapt comment at line 64)

rbovard (changes requested):

permits/api.py
monodo: Actually, this is a mistake: we should filter event which ended before today!

monodo: Fixed

PR #404, opened 2022-02-08
Add some other field types in search

rbovard:
Fix https://jira.liip.ch/browse/YC-607

Please review.

For now, the rendering of multiple list is like that:
image

I don't know if we can improve this rendering without breaking the resumé layout.

### DIFF
--- a/permits/models.py
+++ b/permits/models.py
@@ -55,9 +55,16 @@ ACTOR_TYPE_CHOICES = (
 )
 
 # Input types
-INPUT_TYPE_LIST_SINGLE = "list_single"
+INPUT_TYPE_ADDRESS = "address"
+INPUT_TYPE_CHECKBOX = "checkbox"
+INPUT_TYPE_DATE = "date"
+INPUT_TYPE_FILE = "file"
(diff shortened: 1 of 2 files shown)


rbovard:
Hum, why are the tests failing:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "auth_user_username_key"
DETAIL:  Key (username)=(emiller) already exists.
```

monodo:
Have you tested with, long, very long lists ? Maybe we should crop the result if too long. Moreover I would remove the bold font-weight, no ?

rbovard:
> Maybe we should crop the result if too long

It goes on multiple line, which is not bad?
image
How ever, if the number selected value in the multiple list is too big, the search returns nothing... Any idea why?

> Moreover I would remove the bold font-weight, no ?

I think it's working fine on other field types, it put in evidence where the searched word is.
image

monodo (approved):
thanks @rbovard

rbovard:
Thanks for the merge, but we didn't address this point:

> How ever, if the number selected value in the multiple list is too big, the search returns nothing... Any idea why?

Maybe the rank is to low since the searched value is a little part of the selected values array?

PR #407, opened 2022-02-15
YC-625: Display directives even with incomplete steps

rbovard:
Fix https://jira.liip.ch/browse/YC-625

Please review.


monodo (approved):
thanks!

PR #412, opened 2022-02-22
Missing migration in YC-384

monodo:
@nicolas-sitylb looks like a small change in models META was not included in migrations from YC-384


rbovard (approved):

nicolas-sitylb (approved):
OK

PR #413, opened 2022-02-22
YC-629: Use actor name in API (instead of its absolute ID)

rbovard:
@nicolas-sitylb Please take a look for the plugin part.


monodo:
Require manual check that QGIS expression will work with modified API

rbovard:
It's working fine, I kept the actor typse display to have a 1:1 print result:

Before:
image

After:
image

monodo:
Thanks @rbovard

PR #415, opened 2022-02-22
update django and dependencies

monodo:


rbovard (approved):
Not a dev review but works fine on my instance.

monodo:
NOT DELETING THE BRANCH TO KEEP TRACKING EASY

PR #417, opened 2022-02-23
YC-542: Do not display password change in menu for social accounts

rbovard:
Fix https://jira.liip.ch/browse/YC-542


monodo:
Sorry, merging to fast

PR #426, opened 2022-03-15
YC-643 

AlexandreJunod:
Change the result returned by /rest/current_user


permits/tests/test_permit_request_api.py
```diff
@@ -476,20 +477,36 @@ def test_authorized_network_does_not_raise_exception(self):
         )
         self.assertEqual(response.status_code, 200)
 
-    def test_current_user_returns_correct_username_and_email(self):
+    def test_current_user_returns_users_informations(self):
```
monodo, line 480: one "s" too much ;-)

PR #432, opened 2022-04-01
YC-536 Upload de documents complémentaires par le secrétariat et les services validateurs

kayoumido:


permits/forms.py
```diff
@@ -1322,6 +1322,96 @@ def save(self, commit=True):
...
+        if not self.cleaned_data.get(
+            "authorised_departments"
+        ) and not self.cleaned_data.get("is_public"):
+            raise ValidationError(
+                _(
+                    "Un département doit être renseigner ou le document doit être publique"
+                )
+            )
+
+        if not self.cleaned_data.get("document_type"):
```
danickfort, line 1406: Use the `required` property on this field instead of checking it here

permits/management/commands/fixturize.py
```diff
@@ -602,3 +603,37 @@ def create_template_customization(self):
             application_subtitle="Demandes en lignes",
             application_description="Demandes concernant l' <i>administration</i>",
         )
+
+    def create_document_types(self, wot):
+        document_types = [
+            (
+                "{} Parent #1".format(wot.pk),
```
danickfort, line 610: You could use the `range()` Python built-in function to achieve the document type creation, instead of manually typing the Children 😛

permits/models.py
```diff
@@ -1211,6 +1214,85 @@ class Meta:
...
+    document_type = models.ForeignKey(
+        "ComplementaryDocumentType",
+        null=True,
+        on_delete=models.SET_NULL,
+        verbose_name=_("Type du document"),
+    )
+
+    def delete(self, using=None, keep_parents=False):
+        # delete the uploaded file
+        os.remove(os.path.join(settings.MEDIA_ROOT, self.document.name))
```
danickfort, line 1262: Catch `OSError` exceptions if there is a problem when removing the file, and raise `ProtectedError` with a message explaining why the Document can't be deleted.
https://docs.djangoproject.com/en/3.2/ref/exceptions/#django.db.models.ProtectedError

permits/services.py
```diff
@@ -1230,6 +1230,21 @@ def get_contacts_summary(permit_request):
     return contacts
 
 
+def get_permit_complementary_documents(permit_request, user):
+    groups = user.groups.all()
```
danickfort, line 1234: Declaring variable is superfluous

permits/views.py
```diff
@@ -609,6 +652,57 @@ def handle_prolongation_form_submission(self, form):
...
+    final_error_message = _("Les documents finaux ne peuvent pas être supprimés")
+    owner_error_message = _("Vous pouvez seulement supprimer vos documents")
+
+    def get(self, request, *args, **kwargs):
+        return self.post(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        obj = self.get_object()
+
+        if not obj.owner == self.request.user:
```
danickfort, line 688: I would check with the client with this, but I would let a superuser delete the documents too, with:
`if not obj.owner == self.request.user or self.request.user.is_superuser:`

static/js/permits_complementary_documents.js
```diff
@@ -0,0 +1,80 @@
...
+
+  // update the number of total forms
+  document.getElementById("id_form-TOTAL_FORMS").value = nbDocuments
+
+  // remove the last form
+  let target = [...documents].pop()
+  target.remove()
+})
+
+document.addEventListener('change', (e) => {
```
danickfort, line 56: Could you not add the event listener directly on the document_type select instead of checking that we're working with the correcte element in the next few lines of code?

permits/templates/permits/_permit_request_summary.html
```diff
@@ -109,5 +109,44 @@ <h4>{% trans "Éléments géographiques touchés par la demande (automatique)" %
...
+        <h2>Documents</h2>
+
+        {% for document in documents %}
+          <dl class="mb-4">
+            <dt>{% trans "Nom" %}</dt>
+            <dd>
+                <div class="pl-0 col-6">
+                  {{ document.document }}
+                  {% if document.is_public %}
+                    <span class="badge badge-pill badge-success">{% trans "Publique" %}</span>
```
danickfort, line 124: The pill-badge makes the text look very small in the summary page, maybe make it bigger?

danickfort (dismissed):
After review, a few small adjustments to make. Also, after the sprint-9 stories are merged, these complementary documents might have to be included in the Permits API, in order to display the public documents in the Geo Calendar, but this might be out of scope for this story (please see with client 😄 )

static/js/permits_complementary_documents.js
kayoumido: The problem comes from the extra forms that can be created afterwards. If I use `.addEventListener('click'..)` directly on the select, it won't apply to the selects that get created afterwards

permits/forms.py
```diff
@@ -1322,6 +1322,93 @@ def save(self, commit=True):
...
+        if not self.cleaned_data.get(
+            "authorised_departments"
+        ) and not self.cleaned_data.get("is_public"):
+            raise ValidationError(
+                _(
+                    "Un département doit être renseigner ou le document doit être publique"
+                )
+            )
+
+        if not cleaned_data["parent_{}".format(cleaned_data["document_type"].pk)]:
```
danickfort, line 1406: Crashes if not document type is selected.
`cleaned_data.get("document_type")`

danickfort (changes requested):

permits/forms.py
```diff
@@ -1403,7 +1403,10 @@ def clean(self):
                 )
             )
 
-        if not cleaned_data["parent_{}".format(cleaned_data["document_type"].pk)]:
+        if not self.cleaned_data.get("document_type"):
+            raise ValidationError(_("Un type doit être renseigné!"))
```
danickfort, line 1407: The requirement of `document_type` is already validated at this point, so you can just:

`return cleaned_data`

danickfort (changes requested):

PR #446, opened 2022-04-13
YC-605 Fixes from customer review

danickfort-liip:


permits/services.py
```diff
@@ -269,11 +269,16 @@ def _get_properties_filtered(permit_request, props_filter):
...
+    exclusions = [
+        models.WorksObjectProperty.INPUT_TYPE_FILE,
+    ]
+    if additional_type_exclusions is not None:
+        exclusions += additional_type_exclusions
     return _get_properties_filtered(
         permit_request,
         lambda qs: qs.exclude(
-            input_type__in=[models.WorksObjectProperty.INPUT_TYPE_FILE,]
+            input_type__in=[models.WorksObjectProperty.INPUT_TYPE_FILE,] + exclusions
```
rbovard, line 281: Is not `models.WorksObjectProperty.INPUT_TYPE_FILE` added two times?

permits/filters.py
```diff
@@ -61,7 +61,11 @@ class DepartmentPermitRequestFilterSet(BasePermitRequestFilterSet):
         lookup_expr="icontains", label=_("Auteur de la demande"),
     )
     works_object_types__works_object = django_filters.filters.ModelChoiceFilter(
-        queryset=models.WorksObject.objects.order_by("name"),
+        queryset=models.WorksObject.objects.filter(
+            works_object_types__permit_requests__isnull=False
```
rbovard, line 65: Is it possible to filter `WorksObjet` visible for the current user (= displayed in the table) instead of all `WorksObject` with data application wide?

rbovard, line 65: Is it possible to filer WorksObjet visible for the current user (= displayed in the table) instead of all WorksObject with data application wide?

rbovard (commented):
Thanks @danickfort-liip, I just noticed something with the content of the `WorksObject` combox. Is it possible to take a look?

permits/services.py
danickfort-liip: No, the other one is `INPUT_TYPE_FILE_DOWNLOAD`

rbovard: I mean at line 274

permits/filters.py
danickfort-liip: I will take a look at the logic for retrieving the visible objects for the current user and fix the QuerySet if it is simple enough 👍

PR #450, opened 2022-04-16
YC-605 Dashboard filter object type options fix

danickfort:


rbovard (approved):
Thanks a lot for the improvement!

I've also added the same logic to filter `AdministrativeEntity` and `WorksType`, could you please review?

permits/filters.py
```diff
@@ -56,6 +56,30 @@ class OwnPermitRequestFilterSet(BasePermitRequestFilterSet):
     pass
 
 
+def _get_administrative_entities_queryset_for_current_user(request):
+    return (
+        models.PermitAdministrativeEntity.objects.filter(
+            works_object_types__permit_requests__in=services.get_permit_requests_list_for_user(
```
danickfort, line 62: There is also a relation between `PermitAdministrativeEntity` and permit requests in via the `permit_requests` related name, so line 62 should be:
`permit_requests__in=services.get_permit_requests_list_for_user(`

danickfort (commented):
Small change to have an optimized queryset @rbovard 🙂

rbovard:
Thanks for the tip!

Please merge when tests are green.

PR #451, opened 2022-04-20
Admin little improvements

rbovard:
*Replaces #441*

See commits for details.


permits/admin.py
```diff
@@ -222,21 +224,23 @@ def has_add_permission(self, request):
...
         return obj.socialaccount_set.exists()
 
+    is_sociallogin.admin_order_field = "socialaccount"
     is_sociallogin.short_description = "Social"
 
     def get_readonly_fields(self, request, obj=None):
         # limit editable fields to protect user data, superuser creation must be done using django shell
         if request.user.is_superuser:
             return [
-                "email",
```
monodo, line 231: Sure ? You want to allow superuser to edit email addresses ?

rbovard: Yes, if he cannot who can?

I think the superuser is the right user to be able to do that, no?

For the record, the times we needed to do it, we had to go directly into the database...

AlexandreJunod: @monodo Awaiting PO answer. Is it ok to allow superuser to edit ?

Dev review done, just waiting an answer to merge PR. @rbovard

PR #452, opened 2022-04-21
YC-430 Validation de l'email à la création du compte

kayoumido:


geomapshark/views.py
```diff
@@ -68,28 +106,39 @@ def permit_author_create(request):
 
     if is_valid:
         new_user = djangouserform.save()
+        # email wasn't verified yet, so account isn't active just yet
+        new_user.is_active = False
```
danickfort, line 110: Since we are setting accounts as inactive by default, we should probably have a different error message when the user tries to login before activating the account. Currently "Votre mot de passe et votre nom d'utilisateur ne correspondent pas" is kind of misleading

danickfort (changes requested):

geomapshark/tests/test_register.py
```diff
@@ -118,23 +112,28 @@ def test_user_cannot_register_if_captcha_isnot_filled(self):
...
-            response = self.execute_post_register()
-            self.assertEqual(response.status_code, 200)
-            self.assertFalse(response.context["user"].is_authenticated)
-            self.assertRedirects(response, resolve_url("account_login"))
-            self.assertContains(
-                response,
-                "Votre compte a été créé avec succès! Vous allez recevoir un email pour valider votre email",
-            )
-            self.assertEqual(len(mail.outbox), 1)
+    # if settings.ENABLE_2FA:
```
danickfort, line 115: Remove this instead of commenting ? 😄

danickfort (changes requested):

PR #468, opened 2022-05-06
YC-639: Add WOT property to have only on geotime

rbovard:


monodo (changes requested):

monodo:
Thanks @rbovard 
please update wiki & 1.5 release draft to mention this feature

monodo (approved):

PR #478, opened 2022-05-16
YC-739: Add shortname in list + export

monodo:


rbovard:
Hi @monodo,

Is it possible to display these columns only if the setting "enable geocalendar" is true?

rbovard:
To be coherent with:
https://github.com/yverdon/geocity/blob/575ff2a3209ce9001672d1cc3ef8d60169b78e4b/permits/forms.py?plain=1#L951-L953

permits/tables.py
```diff
@@ -211,6 +210,7 @@ class Meta:
         model = models.PermitRequest
         fields = (
             "id",
+            "shortname",
```
rbovard, line 213: Should also be removed.

PR #487, opened 2022-05-25
YC 780 - show public field for public events

monodo:


permits/serializers.py
```diff
@@ -69,7 +69,19 @@ class Meta:
 
 class WotPropertiesValuesSerializer(serializers.RelatedField):
     def to_representation(self, value):
-        wot_properties = services.get_wot_properties(value, value_with_type=True)
+        current_user = None
+        request = self.context.get("request", None)
+        if request:
+            current_user = request.user
+            basic_authentication = request.session._SessionBase__session_key
```
monodo, line 76: Basic auth is disabled by default.  Looks like you mean session_authentication ?

permits/services.py
```diff
@@ -1693,11 +1694,11 @@ def get_wot_properties(value, value_with_type=False):
...
                 wot = f'{prop["works_object_type__works_object__name"]} ({prop["works_object_type__works_type__name"]})'
-                # List of a lost, to split wot in objects. Check if last wot changed or never assigned, means first iteration
+
+                # List of a list, to split wot in objects. Check if last wot changed or never assigned. Means it's first iteration
                 if property and wot != last_wot:
                     wot_properties.append(property)
                     property = []
-
                     # WOT
                     property.append(
```
monodo, line 1703: No need to check if user is authenticated here ?

monodo (commented):
nice way to solve it! Thanks @AlexandreJunod

permits/services.py
AlexandreJunod: No, if we want to show the public `WOT` without any properties. It only shows public requests, cause the queryset selects only the ones the user can see

PR #488, opened 2022-05-30
Reports generation

olivierdalang:
WIP for report generation.

**WARNING**

This PR includes some rework of the permissions. AuthTokens is not anymore optional, but it now uses django-rest-knox instead of DRF's built in implementation. AuthTokens authentication is only available from a whitelisted range of IPs, as it's meant for internal calls (map/PDF generation) or automation.

It also in changes a little bit how tests are run. As explained, to test generation of PDF/QGIS from other containers, they need to be able to talk to Django's LiveServerTestCase, meaning we can't `docker exec` the tests in the running web container, but instead must `docker run` the tests. I updated the readme as well as CI, theoretically it should run the same things (full test suite twice, once with and once without 2FA, and including coverage).

Progress:
- [x] new `reports` app with Report / RepotLayout models
- [x] corresponding admin config 
- [x] basic set of blocks to compose layouts
- [x] PDF generation as previews
- [x] PDF generation as attachments to permit requests
- [x] rewamp QGIS image generation (isolated in containers instead of QGIS server) 
- [x] basic proof-of-concept based on django-streamfield and wkhtmltopdf
- [x] only accept token authentication from internal calls
  -  [x] merge https://github.com/yverdon/geocity/pull/512
  -  [x] merge https://github.com/yverdon/geocity/pull/513
- [x] isolate PDF generation in containers too (like how it's done for map generation)
- [x] remove old print setup
- [x] improve form layout (see [mockup](https://user-images.githubusercontent.com/1894106/173843450-ffac241a-c01d-45ed-abbe-42caa5821395.png))
- [x] automatically create a default report so users don't need to create a layout to start
- [x] actual implementation of some reports (@monodo and @AlexandreJunod fait par vous, on avait dit au moins permis de fouille)
- [x] basic tests (PDF preview endpoint, attach PDF to request form, but not including the actual PDF rendering, so it does not cover that QGIS rendering actually works)
- [ ] address the many new TODOs in code
- [x] If section order is not set manually, section does not appear after save
- [x] Merge blocs improvements by @AlexandreJunod 
- [x] Fix reports: error on both dev and preprod instance when printing PDF, html pdf only works without map

Requiring particular review
- [ ] double(/triple) check the implementation of token authentication, and ensure the logic that was removed will still work as expected
- [x] linkage between reports/reportlayouts and integrator(s) (not sure what is desired, what at least looks strange is that the field is nullable yet required in the admin and the possibility to define inconsistent links (e.g. a reportlayout to one integrator, and a report to a different integrator) => fixed here: https://github.com/yverdon/geocity/pull/531
- [ ] tests suite in CI and locally

Further improvements (maybe):
- [x] consider replacing django-streamfield by sortable+polymorphic inlines (may be better in terms of UI and code organisation) (initial work in https://github.com/yverdon/geocity/pull/510)
- [ ] async rendering of PDFs (will require a task queue)
- [ ] advanced tests (comparing the contents of the generated PDF to test QGIS rendering)
- [ ] include a predefined QGIS project block (so you choose between presets instead of having to upload your own QGIS project)
- [ ] add ability to attach all reports at once
- [ ] add ability to automatically attach reports at some processing steps (validation...)
- [x] add a way to create tokens from the admin (currently, can only be done by the user itself through the API, which is not convenient)
- [ ] add ability to disable blocks in reports (useful when configuring reports, because otherwise you must delete them altogether if you don't want them to appear)


olivierdalang:
Superseded by https://github.com/yverdon/geocity/pull/533

PR #505, opened 2022-06-08
restrict integrator to own object in admin for report & reportlayout

monodo:


olivierdalang:
there was already a N-M relation between Report and WorksObjectType (`reports = models.ManyToManyField("reports.Report", related_name="worksobjecttypes")`).

We could move it to the Report model so it's more visible

PR #511, opened 2022-06-13
enabled field

monodo:


permits/forms.py
```diff
@@ -1,3 +1,4 @@
+from faulthandler import disable
```
AlexandreJunod, line 1: Delete

AlexandreJunod (changes requested):

permits/forms.py
```diff
@@ -954,13 +956,24 @@ def __init__(self, *args, **kwargs):
...
             filter2 = [
                 el
                 for el in filter1
                 if any(i in el for i in available_statuses_for_administrative_entity)
             ]
+
             self.fields["status"].choices = tuple(filter2)
 
+            # A permit that is approved, rejected or archived cannot have its status changed and author cannot be notified anymore
+            if self.instance.status in [
```
AlexandreJunod, line 968: Instead of listing all not editables status, you can check if it's not in the editable list

`if self.instance.status not in models.PermitRequest.EDITABLE_STATUSES`

PR #512, opened 2022-06-15
Tokens: replace `rest_framework.authtoken` by `django-rest-knox`

olivierdalang:
Since this is quite a sensitive topic, will split changes related to tokens in small PRs so it can be properly reviewed.

This replaces rest_framework.authtoken by django-rest-knox instead. Rationale for the change described here: https://james1345.github.io/django-rest-knox/

I **think** this could be merged independently from the reports branch.

@AlexandreJunod @monodo Please review carefully, there may be side effects, and don't know the test suite well enough yet to know whether we can rely on it for this kind of changes.

(btw, tokens are created via http://localhost:9095/api-tokens/login/)

### DIFF
--- a/permits/admin.py
+++ b/permits/admin.py
@@ -20,8 +20,6 @@ from django.urls import re_path, reverse
 from django.utils.decorators import method_decorator
 from django.utils.translation import gettext_lazy as _
 from django.views.decorators.http import require_POST
-from rest_framework.authtoken.admin import TokenAdmin as BaseTokenAdmin
-from rest_framework.authtoken.models import TokenProxy
 
 from geomapshark import settings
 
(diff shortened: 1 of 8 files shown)


permits/admin.py
```diff
@@ -52,10 +50,6 @@
     "delete_group",
     "see_private_requests",
     # DRF Token authentication
-    "view_tokenproxy",
-    "add_tokenproxy",
-    "change_tokenproxy",
-    "delete_tokenproxy",
```
olivierdalang, line 58: I did not understand what exactly this was good for, I guess we can't just delete it.

permits/admin.py
```diff
@@ -1335,45 +1329,6 @@ def has_background_image(self, obj):
     has_background_image.short_description = "Image de fond"
 
 
-# AuthToken admin
-
-
-class TokenAdmin(BaseTokenAdmin):
```
olivierdalang, line 1341: Unfortunately, the built-in admin for knox is not very usable (because knox doesn't store the key, but only the hash). The suggested way to generate tokens is now an API view :  http://localhost:9095/api-tokens/login/

If an admin panel is needed, we will need to add it.

olivierdalang:
(closing this, will be integrated in https://github.com/yverdon/geocity/pull/488)

PR #539, opened 2022-07-01
Use current_site domain instead of a custom definiton...

nicolas-sitylb:
Use current_site domain instead of a custom definiton for defining the domain on the CustomPasswordResertView() class.

### DIFF
--- a/geomapshark/views.py
+++ b/geomapshark/views.py
@@ -75,10 +75,7 @@ class CustomPasswordResetView(PasswordResetView):
 
     def get_context_data(self, **kwargs):
         context = super(CustomPasswordResetView, self).get_context_data(**kwargs)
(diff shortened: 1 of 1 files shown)


nicolas-sitylb:
@AlexandreJunod 
If possible, could you check that it's working on the preproduction server when reviewing this PR.
Also, there is probably something to change [here](https://github.com/yverdon/geocity/blob/f4b5e9be1945ac5d60904d8bf9e34589bd7c606c/templates/registration/password_reset_email.html#L8) base on the work that has been done [there](https://github.com/yverdon/geocity/blob/f4b5e9be1945ac5d60904d8bf9e34589bd7c606c/templates/registration/emails/email_confirmation.txt#L7).
Thanks.

PR #557, opened 2022-09-05
YC-806 General refactoring 2

olivierdalang:
WIP of the general refactoring effort we discussed about. Will let you know when this is ready for review (already opened the PR mainly to track progress and run CI).

This is based on #552 which should be reviewed and merged first.

Progress
- [x] 06/ Simplify entrypoints
- [x] 07/ Rename demo->dev
  - [x] 07A/ env.demo -> .env.example
  - [x] 07B/ docker-compose-demo.yml 
- [x] 08/ Improve .env settings
  - [x] 08A/ Better document local dev settings
  - [x] 08B/ Remove redudent env vars that can be inferred
  - [x]  08C/ Uusing docker-compose.dev.yml now implies ENV=DEV
- [x] 09/ Further improve staticfiles (remove useless whitenoise wrapper)
- [x] 10/ Improve settings.py  (⚠️ **removed some overrides for DEV, need to check that they are indeed useless**)

Notes:
- 06/, removes entrypoint functionality for DEV (neither migrate, nor fixturize). Not sure if the demo deployment relied on this to reset the database regularly ? If so, the demo deployment should be adapted, and use the same deployment pipeline as prod, with the difference of running the fixturize_demo command every now and then (cronjob ?).
- 08/ potentially, we could even remove the ENV envar, and hardcode ENV=DEV in docker-compose.dev.yml, further reducing the risk of running inconsistent states (e.g. PROD with docker-compose.dev.yml)
- 10/ removes some settings overrides for DEV (to make it as close as possible to PROD), we need to make sure they are indeed usless (SESSION_COOKIE_HTTPONLY, CORS_ALLOW_CREDENTIALS)


monodo (approved):

README.md
```diff
@@ -75,12 +75,9 @@ DEFAULT_SITE=geocity.ch
...
 ```ini
-COMPOSE_FILE=docker-compose.yml:docker-compose-demo.yml
-COMPOSE_PATH_SEPARATOR=:
-ENV=DEV
-DEBUG=True
-DEV_DEPENDENCIES=true
-CLEAR_PUBLIC_SCHEMA_ON_FIXTURIZE=true
+PRIVATE_DOCUMENTS_DIR=C:\some\directory\for\mounting\geocity\private_documents
+ARCHIVE_DIR=C:\some\directory\for\mounting\geocity\archive
+BASE_DOMAIN=localhost
```
monodo, line 78: Please update from main => `DEFAULT_SITE`

geocity/wsgi.py
```diff
@@ -10,9 +10,7 @@
 import os
 
 from django.core.wsgi import get_wsgi_application
-from whitenoise import WhiteNoise
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "geocity.settings")
 
 application = get_wsgi_application()
-application = WhiteNoise(application, root=os.environ["STATIC_FILES_ABSOLUTE_PATH"])
```
monodo, line 18: how are static files served on production then ?

olivierdalang: It seems it's done by `whitenoise.middleware.WhiteNoiseMiddleware` (see http://whitenoise.evans.io/en/stable/#quickstart-for-django-apps)

Seems that wrapping the wsgi app was not necessary in the first place (it's for non-django apps)

Locally at least it still works with that change.

geocity/settings.py
```diff
@@ -598,3 +615,13 @@
         ],
     },
 }
+
+
+def show_toolbar(request):
+    """Shows the debug toolbar when `?DEBUG=true` is your URL and DEBUG is enabled."""
+    return DEBUG and strtobool(request.GET.get("DEBUG", "false"))
```
AlexandreJunod, line 622: Use a condition instead of strtobool to keep the same structure as other settings.
`request.GET.get("DEBUG", "false").lower() == "true"`

Or should we adapt everything to use strtobool, to keep the same logic everywhere

AlexandreJunod (changes requested):

README.md
olivierdalang: Ok

geocity/settings.py
olivierdalang: Ok

AlexandreJunod (approved):

PR #559, opened 2022-09-05
add missing templates for password reset

monodo:


olivierdalang (changes requested):
⚠️ with the refactoring these should now go under apps/core/templates (probably not getting conflicts because it's new files)

rbovard:
@monodo Why did you close it?

PR #562, opened 2022-09-06
YC-865: Do not add approved and rejected statuses for permit with backoffice field always updatable

rbovard:
Fix https://jira.liip.ch/browse/YC-865


rbovard:
@monodo I let you confirm that this is what we were looking for before fixing the tests

rbovard:
@monodo I fixed the issue and played my custom workflows which are working fine. Please review.

PR #573, opened 2022-09-20
fix unauthorized api response on unlogged user

anthonyledda:


geocity/apps/permits/api.py
```diff
@@ -288,7 +288,6 @@ class PermitRequestDetailsViewSet(
 
     throttle_scope = "permits_details"
     serializer_class = serializers.PermitRequestDetailsSerializer
-    permission_classes = [BlockRequesterUserWithoutGroup]
```
AlexandreJunod, line 291: Use `permission_classes = [BlockRequesterUserWithoutGroup]` instead of nothing.

**Why ?** 
This is an easier way to block all open routes directly from one location, if there is a security issue in geocity later.

AlexandreJunod (changes requested):

PR #578, opened 2022-09-27
fix YC-880: archive requests with documents fails

olivierdalang:


geocity/apps/permits/tests/functional/test_complementary_documents.py
```diff
@@ -207,3 +209,15 @@ def test_non_owners_can_not_delete_documents(self):
             messages.ERROR,
         )
         self.assertResponseMessageContains(actual, expected)
+
+    # regression test for YC-880
```
monodo, line 213: Keep this comment here ?

olivierdalang: I don't really have an opinion for that (there does not seem to be a standard practice for that in the codebase... could also put it as a docstring or remove it altogether)

monodo: Looks to me that the test method name is explicit enough

monodo:
fixed in phoenix :-)

PR #579, opened 2022-09-27
YC-806 General refactoring 5 - make archive use a django file field instead of custom code

olivierdalang:
(this is based on https://github.com/yverdon/geocity/pull/577 which should be reviewed and merged first)

The `ArchivedPermitRequest` relied on custom code to store and retrieve the files from the file directory (stored as regular file in directories like `/archive/2927_22.06.2022.09.56.52-demande-de-permis-de-construire`).

This was:
- fragile (files will silently be orphaned if the code that generated the file names changes, which it should as we should use YYYY-MM-DD instead of DD.MM.YYYY)
- not portable (sooner or later, file storage will be managed differently than a plain file system volume, e.g. S3 or similar approach)
- not efficient (files stored uncompressed on the file system take more space)

This fixes the situation by adding a `ArchivedPermitRequest.archive`, a FileField similar to other private storage file fields already used. The files are zipped when creating the archive.

⚠️ Important notes ⚠️  
- bulk download of archive now returns an archive or archives, as otherwise we'd need to uncompress all zips on download, which could be heavy. Is that OK ? If not we can uncompress and recompress all at once, but not sure how many requests that could handle before causing issues.
- this includes zipping all existing archives as part of a migration, which could be slow, and **needs to be carefully tested on a copy of prod data to ensure it works as expected** (otherwise we risk to loose data)
- deletion is not yet handled. I think it's not handled for other file fields either. I think this should be done separately and consistently with other file files


geocity/apps/permits/models.py
```diff
@@ -1752,24 +1754,19 @@ class ArchivedPermitRequest(models.Model):
...
-        return os.path.join(settings.ARCHIVE_ROOT, self.dirname)
+        return self.archive.path
 
     def delete(self, using=None, keep_parents=False):
-        # delete the archive
-        shutil.rmtree(self.path)
-
-        # delete the permit request linked to the archive
+        # TODO: is this really wanted ?!!
         self.permit_request.delete()
```
olivierdalang, line 1765: @monodo Stumbled about this: seems deleting an `ArchivedPermitRequest` cascades to it's `PermitRequest`, which I find very counter-intuitive, and could easily lead to future data-loss. I think we should remove that and put that code in the view that handles it.

monodo: Indeed. So , you would add a button that batch deletes selected permitrequests and associated files, that's right ? I'll fill an issue for that

monodo (approved):

PR #408, opened 2022-02-15
YC-604: Add type name in email subject

rbovard:
Fix https://jira.liip.ch/browse/YC-604

Please review.

PR #586, opened 2022-10-06
YC-883: Add objects list in emails

rbovard:
Fix https://jira.liip.ch/browse/YC-883

I also added the types list in mail subject https://github.com/yverdon/geocity/pull/408 where it was missing.

### DIFF
--- a/geocity/apps/permits/models.py
+++ b/geocity/apps/permits/models.py
@@ -940,6 +940,16 @@ class PermitRequest(models.Model):
             )
         )
 
+    def get_works_object_names_list(self):
+
+        return ", ".join(
+            list(
+                self.works_object_types.all()
+                .values_list("works_object__name", flat=True)
+                .distinct()
+            )
+        )
+
     def archive(self, archivist):
         # make sure the request wasn't already archived
         if ArchivedPermitRequest.objects.filter(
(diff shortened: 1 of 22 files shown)


geocity/apps/permits/models.py
```diff
@@ -940,6 +940,16 @@ def get_works_type_names_list(self):
             )
         )
 
+    def get_works_object_names_list(self):
+
+        return ", ".join(
+            list(
+                self.works_object_types.all()
+                .values_list("works_object__name", flat=True)
```
monodo, line 948: Use new shortname field ?

geocity/apps/permits/services.py
```diff
@@ -1907,6 +1913,10 @@ def get_works_type_names_list(permit_request):
     return permit_request.get_works_type_names_list()
 
 
+def get_works_object_names_list(permit_request):
+    return permit_request.get_works_object_names_list()
+
+
```
monodo, line 1919: Please remove. If you can pass `permit_request `to this method, you can call directly the method on the `permit_request `instance

geocity/apps/permits/models.py
rbovard: > Use new shortname field ?

No, as specified here https://jira.liip.ch/browse/YC-883 we want the object name (we already have the type name in the email subject)

> How does the title gets cropped if many WOT are selected ?

Like the type names in the email subject: concatenated and separated by commas.

geocity/apps/permits/services.py
rbovard: So also for `def get_works_type_names_list(permit_request):` I guess?

rbovard:
@monodo Review addressed, please take a look.

monodo (approved):

PR #589, opened 2022-10-06
YC-610: Render line breaks for multiline text fields

rbovard:
Fix https://jira.liip.ch/browse/YC-610

After:
image

Before:
image


monodo (commented):
nice @rbovard

monodo (approved):

PR #590, opened 2022-10-14
YC-907 (fix): Use default WOT name if shortname is not defined

rbovard:
Fix regression introduced in #584, related to https://jira.liip.ch/browse/YC-907


rbovard:
All green, merging.

PR #600, opened 2022-11-15
Phoenix print

monodo:


geocity/apps/django_wfs3/urls.py
```diff
@@ -3,13 +3,13 @@
...
-    r"permits_point", api_views.SubmissionPointViewSet, "submissions_point"
+    r"submissions_point", api_views.SubmissionPointViewSet, "submissions_point"
 )
 wfs3_router.register(
-    r"permits_line", api_views.SubmissionLineViewSet, "submissions_line"
+    r"submissions_line", api_views.SubmissionLineViewSet, "submissions_line"
 )
 wfs3_router.register(
-    r"permits_poly", api_views.SubmissionPolyViewSet, "submissions_poly"
+    r"submissions_poly", api_views.SubmissionPolyViewSet, "submissions_poly"
```
rbovard, line 14: `poly` means nothing, why not `area`?

sephii: @monodo I have no clue about this, could you answer so i can then merge teh PR?

monodo: Good idea, but out of scope of the refactor

rbovard: Yes but we might as well take advantage of it since we rename the urls?

monodo: `Poly ` is not ideal, but closer to OGC geometry types definition thus, not sure renaming to area is a great idea. Correct name would by `Polygon `or `MultiPolygon`

monodo: Anyway, we'll not rename now, but potentially at the end of refactor if time remains

rbovard: Yes but in this case you should also rename line to polyline ;)

PR #630, opened 2022-11-24
Limit data for 3 months for anonymous user

monodo:


geocity/apps/api/views.py
```diff
@@ -39,21 +40,31 @@ def get_queryset(self):
         This view should return a list of events for which the logged user has
         view permissions or events set as public by pilot
         """
-
         show_only_future = self.request.query_params.get("show_only_future", None)
         starts_at = self.request.query_params.get("starts_at", None)
         ends_at = self.request.query_params.get("ends_at", None)
         administrative_entities = self.request.query_params.get("adminentities", None)
 
+        user_is_trusted = is_user_trusted(self.request.user)
```
AlexandreJunod, line 48: `is_user_trusted = is_user_trusted(self.request.user)`
to keep coherence

AlexandreJunod (changes requested):

geocity/apps/accounts/users.py
```diff
@@ -41,6 +41,11 @@ def get_administrative_entities_associated_to_user_as_list(user):
     )
 
 
+# A trusted user is an authenticated user that is associated at least with one administrative entity
+def is_user_trusted(user):
+    return len(get_administrative_entities_associated_to_user(user)) > 0
```
AlexandreJunod, line 46: Rebase on main and use :` return len(AdministrativeEntity.associated_to_user(user)) > 0`

PR #632, opened 2022-11-24
YC-869: Keep url filters after login on 2FA instance

monodo:
Needs to be refactored for a DRYer solution


rbovard:
Refactoring done, and tested with and without `PREFIX_URL`

PR #637, opened 2022-11-30
YC-975: Allow HTML in additional_information also in backoffice

rbovard:
Fix https://jira.liip.ch/browse/YC-975


monodo (approved):
To be replaced by markdown later

PR #638, opened 2022-12-01
Rename route permit-requests to submissions

rbovard:


rbovard:
@monodo Can you please test on your after the refactoring?

- For the warning, I think is a non positive since we use relative redirect (via a the `replace`), no?
- For the migration error, I think it's not related to this PR
    - `CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0006_change_admin_order, 0006_remove_geom_layer in submissions).`

PR #647, opened 2022-12-09
Remove old refs to permit-request in comments/logs

rbovard:
Something that was forgotten during phoenix.


monodo (approved):
thanks @rbovard !

PR #656, opened 2022-12-20
rename submit button

monodo:


rbovard (changes requested):
As discussed, add a checkbox to enable the submit button instead of renaming it.

PR #678, opened 2023-01-25
YC-936 - Nombre maximal de demandes

danickfort:


geocity/apps/submissions/forms.py
```diff
@@ -202,14 +223,34 @@ def __init__(self, instance, form_categories=None, *args, **kwargs):
...
+        forms_by_category = []
+        disabled_choices = set()
+        for category, forms in sorted(
+            forms_by_category_dict.items(), key=lambda item: slugify(item[0].name)
+        ):
+            forms_list = []
+            for form in forms:
+                form_name = form.name
+                if form.has_exceeded_maximum_submissions():
+                    form_name = f"{form_name} <span class='px-5 text-danger'>{form.max_submissions_message}</span>"
```
AlexandreJunod, line 235: `form_name = form.name` can be removed, only used once.

use form.name after delete `form_name = f"{form.name} <span class='px-5 text-danger'>{form.max_submissions_message}</span>"`

AlexandreJunod (changes requested):

geocity/apps/submissions/forms.py
danickfort: `form_name` is used below on line 237, which is why it stored in a separate object

PR #686, opened 2023-02-02
Feature/ssp 530taiga

AlexandreJunod:


AlexandreJunod:
Adapt danick code to use a button instead of an <a href> and adapt js to make number of submissions and checkbox for terms and conditions work together

monodo (approved):

PR #689, opened 2023-02-09
Feature/readme

AlexandreJunod:


rbovard (dismissed):
Thanks, I've made some changes via a commit (instead to make many code suggestions, it was faster to do so).

Feel free to adapt it.

rbovard:
Question: maybe we should put the "Contribution guideline" after "Setting up full Docker non persistent demo" and "Setting up production instance" (--> people want to try more than contribute?). And it will be just before the "Development tools" paragraph.

What do you think?

AlexandreJunod:
> 

It's a good idea

AlexandreJunod:
@rbovard Good idea, I'll move this just before "Development tools"

PR #694, opened 2023-02-10
YC-1030: Remove number of first geotime form

rbovard:
Fix https://jira.liip.ch/browse/YC-1030


rbovard:
@AlexandreJunod Only the last commit should be merged.

PR #710, opened 2023-02-28
Integrate advanced map plugin

monodo:
New widget for geometry edition for geotime form step with advanced functionnalities like single and multiple select on vector features

- [x] Fix error Uncaught SyntaxError: export declarations may only appear at top level of a module [geocity-wc-map.js:23445](http://localhost:9095/static/libs/js/advanced-geometry-edition/geocity-wc-map.js)
- [x] Fix error Uncaught SyntaxError: redeclaration of const toSendEvent <anonymous> http://localhost:9095/submissions/286/geotime/:610
- [x] In admin the web component config might be attached to AdministrativeEntity or/and Form. 
- [x] Add Model to configure the web component. A simple solution might be to use Django native JSONField and use PrettyJSONWidget for admin: https://pypi.org/project/django-jsoneditor/ . Model must have on FK on AdministrativeEntity et one FK on Form. It also have a FK to integrator group.
- [x] Filter FKs for integrator in admin and add permissions to integrator groups
- [x] Add endpoint for configuration and implement overwrite of Form-related configuration over AdministrativeEntity. If a configuration is attached to Form, it takes precedence over the config attached AdministrativeEntity. configuration
     => first attempt: /forms/mapwidgetconfiguration/<form_id>/
- [x] In admin, let the admin decide which widget to use. 
- [x] Swicht the widget in the Form depending on select configuration: https://github.com/yverdon/geocity/blob/main/geocity/apps/submissions/forms.py#L1198-L1201 
- [x] If new widget is chosen, hide fields in admin related to the original widget. Only administrative entity that allows the selection of one form at a time (radio) can use the new widget and the admin user is informed properly of that limitation OR take a global configuration if multiple forms are selected (?).
-  [x] Define Validation pattern for JSON editor in admin and fine tune jsoneditor admin widget
-  [x] Build js using container to avoid copy pasting js code in repo
-  [x] Create js directory in dev config


geocity/fields.py
```diff
@@ -41,6 +41,15 @@ def media(self):
         )
 
 
+class GeometryWidgetAdvanced(geoforms.OpenLayersWidget):
+    template_name = "advancedgeometrywidget/geometrywidget.html"
+    map_srid = 2056
+
+    @property
+    def media(self):
+        return forms.Media()
```
AlexandreJunod, line 50: Ajouter le JS à ce niveau là, plutôt que dans le template ? (geocity/apps/submissions/templates/advancedgeometrywidget/geometrywidget.html)
```suggestion
        return forms.Media(
                js=(
                       "geocity-wc-map/geocity-wc-map.js",
                       "js/advancegeometrywidget/geometrywidget.js",
                ),
        )
```

geocity/apps/core/static/css/admin/display.css
```diff
@@ -0,0 +1,3 @@
+.hide {
+  display: none;
+}
```
AlexandreJunod, line 3: Already present on `geocity\apps\core\static\css\main.css`

```
.display-none {
  display: none;
}
```

AlexandreJunod (changes requested):

geocity/fields.py
monodo: Good point but loading order seems to cause an issue

geocity/apps/forms/models.py
```diff
@@ -673,6 +673,18 @@ def clean(self):
                             )
                         }
                     )
+        # FIXME
+        if (
+            self.geo_widget_option == self.GEO_WIDGET_ADVANCED
+            and not AdministrativeEntity.is_single_form_submissions
```
AlexandreJunod, line 679: Select administrative entity for this form

AlexandreJunod (changes requested):

Issue #525, opened 2022-06-22
We miss a default Site (settings.SITE_ID should not be None)


olivierdalang:
Currently, SITE_ID is set to None in settings, meaning that unless the request matches an existing Site object, we get an error.

This is problematic, because there can be multiple hosts that should refer to the default site site (localhost, 127.0.0.1, web...), so we should not create an entry for each one.

IMO we should have a SITE_ID set, an use a custom `CurrentSiteMiddleware` that overrides the current site if there is a match, and keeps the default otherwise.

(while at it, we could probably also remove the `SetCurrentSiteMixin` mixin, which does not seem alright)

monodo:
Moved to Jira @olivierdalang

PR #715, opened 2023-03-01
yc 646

monodo:


geocity/apps/accounts/templates/registration/password_reset_confirm.html
```diff
@@ -0,0 +1,25 @@
...
+{% load static %}
+{% load i18n %}
+
+{% block content %}
+
+{% if validlink %}
+
+<h2>Geocity</h2>
+<hr>
+<h3>{% translate "Choisissez un nouveau mot de passe!" %}</h3>
```
rbovard, line 12:
```suggestion
<h3>{% translate "Choisissez un nouveau mot de passe" %}</h3>
```

rbovard (changes requested):
I think the first commit was already merged in another PR.

I know it's outside the scope of the story, but is it possible to edit email subjet? Because the current one is not very clear `Password reset on default site`

geocity/apps/accounts/templates/registration/password_reset_confirm.html
rbovard: And on the default site?

monodo: The default site should have a name...but another issue related to default site is opened

monodo: https://github.com/yverdon/geocity/issues/525

rbovard (approved):

PR #725, opened 2023-03-09
fix field translation

monodo:


monodo:
> Make migration

No, this model is not a DB model ;-)

PR #759, opened 2023-03-22
YC 1063 - edit validations

monodo:
- [x] add route
- [x] add template
- [x] add view
- [x] configure modelformset_factory
- [x] add save and save+continue buttons
- [x] add permission for this
- [x] place CTA correctly
- [x] set readonlyfields
- [x] write tests


geocity/apps/submissions/models.py
```diff
@@ -271,6 +271,7 @@ class Meta:
         permissions = [
             ("read_submission", _("Consulter les demandes")),
             ("amend_submission", _("Traiter les demandes")),
+            ("edit_submission_validations", _("Modifier les validations")),
```
rbovard, line 274: Be consistent with `edit_submission` (only use one of these terms `Modifier` (my preference) or `Editer`)

PR #762, opened 2023-03-28
yc-1070 checkbox login

AlexandreJunod:


geocity/apps/accounts/forms.py
```diff
@@ -166,6 +167,14 @@ def __init__(self, *args, **kwargs):
         ),
     )
     captcha = CaptchaField(required=True)
+    policy = forms.BooleanField(
+        required=True,
+        label=mark_safe(
+            _(
+                'J\'ai lu et j\'accepte la <b><a href="https://geocity.ch/legal/confidentialite.pdf" target="_blank">politique de confidentialité</a></b>'
```
rbovard, line 174: You need to use the Constance value (like in the footer), not hardcode it

monodo: +1

monodo (dismissed):

geocity/apps/accounts/forms.py
```diff
@@ -171,7 +172,7 @@ def __init__(self, *args, **kwargs):
         required=True,
         label=mark_safe(
             _(
-                'J\'ai lu et j\'accepte la <b><a href="https://geocity.ch/legal/confidentialite.pdf" target="_blank">politique de confidentialité</a></b>'
+                f'J\'ai lu et j\'accepte la <b><a href="{config.PRIVACY_POLICY_URL}" target="_blank">politique de confidentialité</a></b>'
```
rbovard, line 175: What happens if not defined?

https://github.com/yverdon/geocity/blob/develop/geocity/apps/core/templates/base_generic.html#L220-L223

And are you sure you want to put the link in bold?

AlexandreJunod: It opens the same page again. In bold it's easier to understand this is a link, and shows how important it is

PR #793, opened 2023-04-20
Rename forgotten permit_request variables

rbovard:
Not sure if it's also recommended to rename [here](https://github.com/yverdon/geocity/blob/main/geocity/apps/submissions/shortcuts.py#L14) (since `submission` is already set)?


monodo (approved):
thanks @rbovard !!!

PR #796, opened 2023-04-26
Feature/default admin values

monodo:


rbovard:
Why did you close it? It made sense to change these default values.

PR #799, opened 2023-04-28
YC-1092: Add selected price to dashboard and excel export

rbovard:
Fix https://jira.liip.ch/browse/YC-1092

### DIFF
--- a/geocity/apps/submissions/tables.py
+++ b/geocity/apps/submissions/tables.py
@@ -140,6 +140,12 @@ class GenericSubmissionTable(ColumnShiftTable):
         verbose_name=_("Entité administrative"),
         orderable=False,
     )
+    selected_price = tables.TemplateColumn(
+        verbose_name=_("Tarif"),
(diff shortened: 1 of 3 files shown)


geocity/apps/submissions/tables.py
```diff
@@ -140,6 +140,12 @@ class GenericSubmissionTable(ColumnShiftTable):
         verbose_name=_("Entité administrative"),
         orderable=False,
     )
+    selected_price = tables.TemplateColumn(
+        verbose_name=_("Tarif"),
+        template_name="tables/_submission_selected_price.html",
+        attrs=ATTRIBUTES,
+        orderable=False,
+    )
```
monodo, line 148: What happens is no payment is configured ? Should not this field only be visible when a payment is configured.

rbovard: It's displayed as `—`, like for the submissions without dates:
image

rbovard: Okay for you @monodo?

PR #814, opened 2023-05-12
Feature/yc 1093

monodo:


rbovard (changes requested):
It's not working when you create a new submission:
1. Select admin entity
2. `No Submission matches the given query.`

Are you sure it's the good approach to filter the dashboard?

For me it will be better if we consider a submission as incomplete if the form is not selected (so the user cannot send it, but can come back to step "Objects" to choose the form).

PR #818, opened 2023-05-16
prevent admin to put user in to validator group for same entity

monodo:


geocity/apps/accounts/admin.py
```diff
@@ -109,6 +113,17 @@ def clean_groups(self):
...
+        validator_number = groups.filter(permit_department__is_validator=True).count()
+        validator_number_for_same_entity = (
+            groups.filter(permit_department__is_validator=True)
+            .distinct("permit_department__administrative_entity")
+            .count()
+        )
+        if validator_number > validator_number_for_same_entity:
+            raise forms.ValidationError(
+                MULTIPLE_VALIDATOR_FOR_SAME_ENTITY_ERROR_MESSAGE
+            )
```
AlexandreJunod, line 126: It's checking for the group and not the actual user, it shouldn't work

AlexandreJunod (changes requested):

AlexandreJunod:
https://jira.liip.ch/browse/YC-665

AlexandreJunod (approved):

geocity/apps/accounts/admin.py
AlexandreJunod: Done

PR #830, opened 2023-06-05
Feature/add missing prefetch for public inquiry

monodo:


monodo:
replaced by https://github.com/yverdon/geocity/pull/833

PR #843, opened 2023-06-19
Excel export with new api

AlexandreJunod:
### Updates

Improved the data returned by the export bouton on the dashboard.
There's more informations, more security and less actions to reach the same result as before.

Here is a quick overview of the changes : 

- Use api updates (serializers), cause there are breaking changes
- Use pandas to display all son levels
- Sort selected_forms on different sheets to avoid having too many elements in first line (title)
- No need to filter entity, form and category to see the fields on the export


geocity/apps/api/serializers.py
```diff
@@ -42,150 +43,181 @@ def get_form_fields(
...
-                    or (current_user and prop["submission__author"] == current_user.id)
-                    or prop[
+                    or (current_user and field["submission__author"] == current_user.id)
+                    or field[
                         "field_values__field__is_public_when_permitrequest_is_public"
                     ]
                 ):
                     # get_property_value return None if file does not exist
-                    file = get_field_value_based_on_field(prop)
+                    file = get_field_value_based_on_field(field)
```
monodo, line 121: Out of scope, but this might be a performance leak as it calls DB on each loop:
```
def get_field_value_based_on_field(prop):
    property_object = FieldValue.objects.get(
        field_id=prop["field_values__field_id"],
        selected_form_id=prop["id"],
    )

    return property_object.get_value()
```

We should make on call on DB for all Field Values and then manipulate them, ideally using vetorized operations, as we now have 🐼 onboard.

geocity/apps/api/serializers.py
```diff
@@ -42,150 +43,181 @@ def get_form_fields(
...
+                    # Put the title
+                    "title": {
+                        "form": field["form__name"],
+                        "category": field["form__category__name"],
+                        "form_category": form_category,
+                    },
+                    "fields": {
+                        field["field_values__field__api_name"]: {
+                            "name": field["field_values__field__name"],
+                            "value": get_field_value_based_on_field(field).url,
```
monodo, line 171: Same comment as above regarding perfomance

geocity/apps/api/serializers.py
```diff
@@ -42,150 +43,181 @@ def get_form_fields(
...
+                        }
+                        # Check this is a file and the file exist
+                        if field["field_values__field__input_type"] == "file"
+                        and get_field_value_based_on_field(field)
```
monodo, line 175: idem

geocity/apps/forms/models.py
```diff
@@ -694,6 +701,20 @@ def clean(self):
                 }
             )
 
+        if self.api_name:
+            if self.api_name != convert_string_to_api_key(self.api_name):
+                print(self.api_name)
+                print(convert_string_to_api_key(self.api_name))
+                raise ValidationError(
```
monodo, line 708: remove `print`

geocity/apps/forms/models.py
```diff
@@ -762,6 +783,12 @@ class Field(models.Model):
         limit_choices_to={"permit_department__is_integrator_admin": True},
     )
     name = models.CharField(_("nom"), max_length=255)
+    api_name = models.CharField(
+        _("Nom dans l'API"),
+        max_length=255,
+        blank=True,
+        help_text=_("Se génère automatiquement lorsque celui-ci est vide."),
+    )
```
monodo, line 791: Should be unique within a form in order no avoid duplicate keys in json output, no ?

geocity/apps/forms/models.py
```diff
@@ -895,3 +922,17 @@ def clean(self):
         if self.input_type == INPUT_TYPE_REGEX:
             if not self.regex_pattern:
                 raise ValidationError({"regex_pattern": _("This field is required.")})
+
+        if self.api_name:
+            if self.api_name != convert_string_to_api_key(self.api_name):
+                print(self.api_name)
+                print(convert_string_to_api_key(self.api_name))
```
monodo, line 929: remove `print`

geocity/apps/submissions/models.py
```diff
@@ -1727,6 +1728,12 @@ class Meta:
 
 class SubmissionAmendField(models.Model):
     name = models.CharField(_("nom"), max_length=255)
+    api_name = models.CharField(
+        _("Nom dans l'API"),
+        max_length=255,
+        blank=True,
+        help_text=_("Se génère automatiquement lorsque celui-ci est vide."),
+    )
```
monodo, line 1736: Should be unique to avoid within a form in order to avoid duplicate key in json ?

geocity/apps/submissions/models.py
```diff
@@ -1757,6 +1764,21 @@ class Meta:
     def __str__(self):
         return self.name
 
+    def clean(self):
+        if self.api_name:
+            if self.api_name != convert_string_to_api_key(self.api_name):
+                print(self.api_name)
+                print(convert_string_to_api_key(self.api_name))
```
monodo, line 1771: remove `print`

geocity/apps/submissions/tables.py
```diff
@@ -363,3 +371,67 @@ class Meta:
...
+        submissions_qs = self.get_table_data().filter()
+
+        # # Make sure there will be no bypass
+        # submissions_list = submissions_qs.values_list("id", flat=True)
+        # visible_submissions_for_user = models.Submission.objects.filter_for_user(
+        #     self.request.user,
+        # ).values_list("id", flat=True)
+
+        # if not all(item in visible_submissions_for_user for item in submissions_list):
+        #     raise SuspiciousOperation
```
monodo, line 392: WIP

geocity/apps/submissions/views.py
```diff
@@ -1615,7 +1615,7 @@ def submission_media_download(request, property_value_id):
 @method_decorator(login_required, name="dispatch")
 @method_decorator(check_mandatory_2FA, name="dispatch")
 @method_decorator(permanent_user_required, name="dispatch")
-class SubmissionList(ExportMixin, SingleTableMixin, FilterView):
+class SubmissionList(PandasExportMixin, SingleTableMixin, FilterView):
```
monodo, line 1618: I guess we should leave a option to the user to use the "old" excel export, thus adapt the mixing dynamically depending on a specific action ?

monodo (changes requested):

geocity/apps/submissions/views.py
AlexandreJunod: Old excel export is still usefull ?

geocity/apps/api/serializers.py
AlexandreJunod: Only used when it's a file

geocity/apps/api/serializers.py
AlexandreJunod: Only used when it's a file

geocity/apps/submissions/templates/submissions/submission_toolbox_header.html
```diff
@@ -30,11 +30,19 @@
...
           <i class="fa fa-file"></i>
-          {% translate "Exporter" %}
+          {% translate "Export simple" %}
         </a>
     </div>
+    {% if user_is_backoffice_or_integrator %}
+    <div class="p-2 permits-action-btns">
+      <a class="btn btn-outline-primary" href="{% url 'submissions:submissions_list' %}?{{ export_csv_url_params_advanced }}" role="button">
+        <i class="fa fa-file"></i>
+        {% translate "Export avancé" %}
```
monodo, line 42: "Export complet" instead of "Export simple" ?

monodo (changes requested):
Almost there 👍

geocity/apps/submissions/tables.py
```diff
@@ -361,3 +374,80 @@ class Meta:
...
+        entities = AdministrativeEntity.objects.associated_to_user(self.request.user)
+
+        # Take all submission except status draft
+        submissions_qs = self.get_pandas_table_data().filter(
+            Q(administrative_entity__in=entities),
+            ~Q(status=models.Submission.STATUS_DRAFT),
+        )
+
+        if not has_permission_to_amend_submission(
+            self.request.user, submissions_qs.first()
```
AlexandreJunod, line 399: `is_backoffice_of_entity` use this instead ? Does the submission really need to be amendable ? @monodo

monodo (approved):

PR #846, opened 2023-06-26
improve recipient to select every type of contact

AlexandreJunod:
### Updates

Improve recipient section to use any type of contact and author. It is possible now to select every type and define a principal and secondary choice.
Data migrations handles the last configuration, so nothing breaks for existing recipient sections


geocity/tests/reports/test_reports.py
```diff
@@ -302,15 +302,6 @@ def test_block_gallery(self):
                     "padding_top": 20,
                     # "is_new_page": bool,
                     "is_recommended": True,
-                    "uses_dynamic_recipient": True,
-                }
-            },
-            20: {
-                SectionRecipient: {
-                    "padding_top": 50,
-                    # "is_new_page": bool,
```
monodo, line 311: remove comment ?

geocity/apps/reports/models.py
```diff
@@ -650,40 +651,72 @@ class Meta:
...
+            )
+            in contacts
+        ):
+            recipient = contacts[
+                convert_string_to_api_key(
+                    dict(CONTACT_TYPE_CHOICES_RECIPIENT)[self.secondary_recipient]
+                )
+            ]
+        else:
+            recipient = "Aucun contact trouvé"
```
monodo, line 712: Will that print "Aucun contact trouvé" in pdf 🤔 ?

geocity/apps/reports/models.py
```diff
@@ -650,40 +651,72 @@ class Meta:
...
     )
-    uses_dynamic_recipient = models.BooleanField(
-        _("Destinataire dynamique"),
-        default=True,
+    principal_recipient = models.PositiveSmallIntegerField(
+        _("Destinataire principal"),
+        choices=CONTACT_TYPE_CHOICES_RECIPIENT,
+        default=CONTACT_TYPE_AUTHOR,
+        help_text=_(
+            "Utilisé par défaut. Si celui-ci n'existe pas, prend le destinataire secondaire"
```
monodo, line 672: There will always be an author and thus a principal recipient, isn't it ?

monodo (changes requested):

geocity/apps/reports/models.py
AlexandreJunod: If this isn't for anonymous submissions, yes

monodo (approved):

PR #857, opened 2023-08-02
Feature YG-1071 enumeration-attack-proof email signup

maltherd:
- [x] Update django-axes
- [x] Adapt django-axes settings in order to avoid unwanted locked users
- [x] ⚠️ Adapt .env config


geocity/apps/accounts/auth_backends.py
```diff
@@ -0,0 +1,20 @@
...
+
+class EmailAuthenticationBackend(ModelBackend):
+    def authenticate(self, request, email, password, **kwargs):
+        User = get_user_model()
+
+        if not email or not password:
+            return None
+
+        try:
+            user = User.objects.get(email=email)
```
maltherd, line 13: This fails when database contains multiple users with same email.

Should we transform it to `User.objects.filter(email=email).first()` to avoid errors, or make some kind of a data migration to set duplicate emails to NULL and send a warning email?

Anyway, the email field of the User model cannot be changed to `unique=True`, since a custom User model was not used.

I'm waiting on a policy choice :)

monodo:
@maltherd the easy login buttons do no longer work locally, could yo take a look ? I think you need to adapt the element name in the js `login `function.

monodo:
@maltherd @rbovard  should we not display (readonly ?)  the username (if exists) and the email in the `/profile/edit/` view ?
image

monodo:
@maltherd rename the placeholder to "Email ou Identifiant" in order to avoid confusion by user who might try to write john@doe.com/johny ?
image

rbovard:
> @maltherd @rbovard should we not display (readonly ?) the username (if exists) and the email in the `/profile/edit/` view ?

Of course, even editable like now, no? Because how can I update my email if I want to?

rbovard:
> @maltherd rename the placeholder to "Email ou Identifiant" in order to avoid confusion by user who might try to write [john@doe.com](mailto:john@doe.com)/johny ?

+1 I would suggest "Email ou identifiant" (with `i`)

rbovard:
> Only email, right ? Don't display username if None and read username readonly, as we don't want to use it in the future, right ?

+1

And whic is displayed on the top right menu?

maltherd:
> @maltherd the easy login buttons do no longer work locally, could yo take a look ? I think you need to adapt the element name in the js `login `function that is called upon click on the role login buttons:
> 
> image

The js login function is provided in the `APPLICATION_DESCRIPTION` constance setting which is provided by the `fixturize` management command. The issue is due to the field changing name (`id_username` → `id_email_or_username`). 

The `fixturize` code was patched ([see here](https://github.com/yverdon/geocity/pull/857/files#diff-2ed08812d93877a9192dd7d206072cc9b5d02edae0e271ab4cef5c5c03431ff0R677)) but for existing installations, the only choice (I think) is to rerun fixturize or to change the constance setting. There aren't constance migrations, are there ?

maltherd:
> > Only email, right ? Don't display username if None and read username readonly, as we don't want to use it in the future, right ?
> 
> +1
> 
> And which value is displayed on the top right menu?

In this PR's code, the full name (first name + last name) is currently displayed in the top right [(see here)](https://github.com/yverdon/geocity/pull/857/files#diff-5ba75ae233a2d2faeb764c4711b188f847b4521267af613544521fc1aae96c01R112)

geocity/apps/accounts/auth_backends.py
maltherd: I did the `first()` then :)

maltherd:
Solved comments have been reacted to with :+1:

monodo:
@maltherd it looks like with the new login system, tested on preprod instance, Django-Axes lockout configuration `AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP `(https://django-axes.readthedocs.io/en/latest/4_configuration.html ) does not manage to discriminate between users and locks any account coming from the same IP address after 3 failing attemps. 
Would you adapt the login system or configure Django-Axes differently ?

PR #858, opened 2023-08-03
update all

monodo:


monodo:
@rbovard https://github.com/pfpayments/python-sdk update to v4.0 please check everthing works on you side.

rbovard:
Do you have a changelog between v3.4 and v4.0 since it can have some API breaks?

rbovard (approved):
I've tested my workflow with payment and it works well.

(But I didn't test all the application)

PR #859, opened 2023-08-03
YC-1136

AlexandreJunod:
Remake contacts to use the db instead of being an enum.

- Rename ContactType by ContactForm. Type is used to define the ContactType of a contact, ContactForm is used to define which contact is used in witch form_category.
- Adapt method and field names to use form instead of type
- ContactType can be added by admin without changing any code
- User can add new contacts multiple times based on integrator configuration
- API handles multiple times the same type of contact

TODO : 

- [x] Adapt reports
- [x] Check if "Auteur" isn't a problem in the contacts list, as it is used only to print correctly in reports and can be avoided
- [ ] Check every variable/method/field to be sure the name is appropriate
- [x] Check every migration is working correctly


geocity/apps/reports/models.py
```diff
@@ -742,9 +740,8 @@ def _get_recipient(self, request, base_context):
             if recipient == 999 and "author" in properties:
                 selected_recipient = properties["author"]
             else:
-                contact = convert_string_to_api_key(
-                    dict(CONTACT_TYPE_CHOICES_RECIPIENT)[recipient]
-                )
+                # TODO: Adapt this part
+                contact = None
```
AlexandreJunod, line 744: TODO

geocity/apps/submissions/templates/submissions/submission_contacts.html
```diff
@@ -28,14 +28,26 @@ <h4>Adresse de facturation si différente de celle de l'auteur ({{ userprofile.f
...
+              <input id="self_contact_{{ forloop.counter }}" type="checkbox" onclick="update_form_value( '{{ forloop.counter0 }}', {{ userprofile }} )" />
+              <label for="self_contact_{{ forloop.counter }}">{% translate "Reprendre mes informations" %}</label>
+              {{ form.media }}
+              {% bootstrap_form form layout='horizontal' %}
+            </div>
+          </div>
+          <hr>
+        </div>
+      {% endfor %}
+      {% if has_any_dynamic_contacts_forms %}
```
monodo, line 44: hide / inform user if max is reached

geocity/apps/reports/models.py
```diff
@@ -739,14 +738,18 @@ def _get_recipient(self, request, base_context):
         selected_recipient = ""
 
         for recipient in recipients:
-            if recipient == 999 and "author" in properties:
+            if recipient.name == "Auteur" and "author" in properties:
```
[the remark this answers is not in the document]
AlexandreJunod: Done

monodo (approved):

PR #861, opened 2023-08-07
YG-Y33 Add short URL parameter to access a form

maltherd:


maltherd:
Todo : add a test

geocity/apps/submissions/steps.py
```diff
@@ -83,6 +83,13 @@ def get_selectable_categories(submission=None, category_tags=None):
     ).distinct()
 
 
+def get_selectable_form(user, quick_access_slug):
+    forms = models.Form.objects.filter(quick_access_slug=quick_access_slug)
+    if not user.has_perm("submissions.view_private_form"):
+        forms = forms.filter(is_public=True)
```
monodo, line 89: @AlexandreJunod I think there is one filter to add here, do you confirm ?

geocity/apps/submissions/steps.py
```diff
@@ -83,6 +83,13 @@ def get_selectable_categories(submission=None, category_tags=None):
     ).distinct()
 
 
+def get_selectable_form(user, quick_access_slug):
+    forms = models.Form.objects.filter(quick_access_slug=quick_access_slug)
```
monodo, line 87: Do we ensure a slug is unique and related to a unique form ?

geocity/apps/submissions/views.py
```diff
@@ -1141,6 +1134,25 @@ def submission_select_administrative_entity(request, submission_id=None):
...
+
+        if form:
+            entities = form.administrative_entities.all()
+
+            if len(entities) == 1:
+                submission = models.Submission.objects.create(
+                    administrative_entity=entities[0],
+                    author=request.user,
+                )
+                submission.forms.set([form])
```
monodo, line 1151: If the same slug is defined on multiple forms of the same entity, multiple forms can be selected with a slug, is that correct and wished @rbovard ?

monodo (commented):
@maltherd just a few comments and question, look good otherwise. @rbovard will check also.

geocity/apps/submissions/steps.py
maltherd: Yes, as can be seen in the diff of geocity/apps/forms/models.py, the slug column is marked `unique`

geocity/apps/submissions/views.py
maltherd: Just in case, since slugs are stored in a `unique` column of the Form table, the "if" you mention here cannot be realized: only one form may have a given slug at a time.

geocity/apps/submissions/views.py
```diff
@@ -1141,6 +1134,25 @@ def submission_select_administrative_entity(request, submission_id=None):
         else None
     )
 
+    quick_access_slug = request.GET.get("q")
```
rbovard, line 1137: The param shoud be named "form"

geocity/apps/forms/models.py
```diff
@@ -518,6 +520,14 @@ class Form(models.Model):
         related_name="map_widget_configuration_form",
         verbose_name=_("Configuration de la carte avancée"),
     )
+    quick_access_slug = models.TextField(
+        blank=True,
+        default=uuid.uuid4,
+        unique=True,
+        validators=[validate_slug],
+        verbose_name=_("URL courte"),
+    )
```
rbovard, line 529: Could you please add (after rename `q` to `form` of course):
`help_text="Permettant d'accéder directement au formulaire par l'url: https://geocity.ch/?form=demande-macaron",`

rbovard (changes requested):
Thanks a lot for this work, it's working fine!

Please take a look to my inline review.

geocity/apps/forms/migrations/0023_add_form_quick_access_slug.py
```diff
@@ -0,0 +1,33 @@
...
+            model_name="form",
+            name="quick_access_slug",
+            field=models.TextField(
+                blank=True,
+                default=uuid.uuid4,
+                null=True,
+                validators=[
+                    django.core.validators.RegexValidator(
+                        re.compile("^[-a-zA-Z0-9_]+\\Z"),
+                        "Enter a valid “slug” consisting of letters, numbers, underscores or hyphens.",
```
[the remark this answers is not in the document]
maltherd: This is because I used [validate_slug](https://docs.djangoproject.com/en/4.2/ref/validators/#validate-slug), a django-provided validator which evaluates to the RegexValidator you can see in the migration. 

So, the validation text is defined in the django source, and worry not, since it is localized ([proof here](https://github.com/django/django/blob/main/django/core/validators.py#L259)) and appears in French if that is the browser language (I tested this).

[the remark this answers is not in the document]
rbovard: Thanks for the feedback!

geocity/apps/forms/models.py
rbovard: Just add the ":" before the example to be consitant with the other help texts.

And should be also in the migration?

maltherd: Yep, sorry for the oversight :+1:

rbovard (approved):
Thanks a lot @maltherd, it's all good for me.

@AlexandreJunod could you please check @monodo comment https://github.com/yverdon/geocity/pull/861#discussion_r1293052048

geocity/apps/submissions/steps.py
AlexandreJunod: What do you mean by one filter ? If user cannot view private forms, we should show only the ones in public that correspond to the quick_access_log.

PR #867, opened 2023-09-06
Anonymous issues and features

AlexandreJunod:
adapt quick_access_slug to work with anonymous forms, prevent from accessing anonymous forms without being anonymous user

### DIFF
--- a/geocity/apps/accounts/users.py
+++ b/geocity/apps/accounts/users.py
@@ -51,10 +51,19 @@ def get_integrator_permissions():
     )
 
 
-def get_users_list_for_integrator_admin(user):
+def get_users_list_for_integrator_admin(user, remove_anonymous=False):
     # Integrators can only view users for restricted email domains.
     if user.is_superuser:
-        return User.objects.all()
+        qs = User.objects.all()
+
+        # Used to remove anonymous users from the list
+        anonymous_users = []
+        for user in qs:
+            if remove_anonymous and user.userprofile.is_anonymous:
+                anonymous_users.append(user.pk)
+        qs = qs.exclude(pk__in=anonymous_users)
+
+        return qs
 
     user_integrator_group = user.groups.get(permit_department__is_integrator_admin=True)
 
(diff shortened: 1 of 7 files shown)


geocity/apps/submissions/views.py
```diff
@@ -1120,6 +1084,99 @@ def anonymous_submission(request):
             return redirect(step.url)
 
 
+def get_anonymous_submission_by_tags(request, entityfilter, typefilter):
+    # Validate entity
+    entities = AdministrativeEntity.objects.public().filter_by_tags(entityfilter)
+    if len(entities) != 1:
+        raise Http404
```
monodo, line 1091: We should inform the integrator that for anonymous submission configuration, tags on administrative entities must be unique, no ?

geocity/apps/submissions/views.py
```diff
@@ -1120,6 +1084,99 @@ def anonymous_submission(request):
...
+    form_categories = (
+        models.FormCategory.objects.filter(forms__is_anonymous=True)
+        .filter_by_tags(typefilter)
+        .distinct()
+        .values_list("pk", flat=True)
+    )
+
+    if len(form_categories) != 1:
+        raise Http404
+    form_category = form_categories[0]
```
monodo, line 1108: idem

geocity/apps/submissions/views.py
```diff
@@ -1120,6 +1084,99 @@ def anonymous_submission(request):
...
+        raise Http404
+
+    if anonymous_selected_forms:
+        entities = anonymous_selected_forms.administrative_entities.all()
+    else:
+        raise Http404
+
+    # Validate entity
+    if len(entities) != 1:
+        raise Http404
```
monodo, line 1134: idem, we should find a way to avoid such unpossible config

geocity/apps/submissions/views.py
AlexandreJunod: It's done like that because it's unique with `entityfilter` `typefilter`. They don't want to block "urb; pdc; sports; ssp".

And informing is a bad approach, when we could just restrict in the model as unique, and force it to be unique.

geocity/apps/accounts/users.py
```diff
@@ -51,10 +51,19 @@ def get_integrator_permissions():
...
+def get_users_list_for_integrator_admin(user, remove_anonymous=False):
     # Integrators can only view users for restricted email domains.
     if user.is_superuser:
-        return User.objects.all()
+        qs = User.objects.all()
+
+        # Used to remove anonymous users from the list
+        anonymous_users = []
+        for user in qs:
+            if remove_anonymous and user.userprofile.is_anonymous:
```
monodo, line 62: Move the `remove_anonymous` if test above the `for `loop in order to avoid looping when not needed ?

PR #881, opened 2023-10-06
Use same term everywhere for integrator

rbovard:


rbovard:
@monodo One question: Why not rename it to "Groupe des intégrateurs" to be more clear? Since the group has a property "Intégrateur".

monodo (approved):

PR #920, opened 2024-01-18
Sort entities in fixtures

rbovard:


rbovard:
@monodo I don't understand why `test_fields_step_order_fields_for_existing_submission` failed. Any idea?

PR #933, opened 2024-02-05
run collect static in anycase

monodo:
⚠️ If migrations are not run, `update_integrator_permissions` must also be run automatically

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -7,17 +7,18 @@ set -e
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
 # On PROD, we run migrations at startup unless explicitly disabled.
-# If disabled, these commands must be executed manually.
+# If disabled, this command must be run manually for the application to function correctly after a model update.
 if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "true" ]; then
     python3 manage.py migrate
-    python3 manage.py update_integrator_permissions
 fi
 
-# On PROD we always collect statics
+# On PROD, we always collect statics
 if [ "$ENV" == "PROD" ]; then
-    python3 manage.py compilemessages -l fr
     python3 manage.py collectstatic --no-input
 fi
+
+python3 manage.py update_integrator_permissions
+python3 manage.py compilemessages -l fr
 
 # Run the command
 exec $@
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -6,9 +6,17 @@ set -e
 # If not will, the container will fail and restart.
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
-# On PROD, we run migrations at startup unless explicitly disabled.
-if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "true" ]; then
-    scripts/migrate.sh
+# On PROD, we run migrations at startup unless explicitly disabled. If disabled,
+# this commands must be run manually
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -6,9 +6,17 @@ set -e
 # If not will, the container will fail and restart.
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
-# On PROD, we run migrations at startup unless explicitly disabled.
-if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "true" ]; then
-    scripts/migrate.sh
+# On PROD, we run migrations at startup unless explicitly disabled. If disabled,
+# this commands must be run manually
+if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "false" ]; then
```
rbovard, line 11: Why did you change that to `"${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "false"`?

I disabled the `migrate.sh` and now you simply want to disable the `migrate` and `update_integrator_permissions`, not change the logic.

rbovard, line 11:
```suggestion
if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "true" ]; then
```

rbovard (changes requested):

entrypoint.sh
```diff
@@ -6,9 +6,17 @@ set -e
 # If not will, the container will fail and restart.
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
-# On PROD, we run migrations at startup unless explicitly disabled.
+# On PROD, we run migrations at startup unless explicitly disabled. If disabled,
+# this commands must be run manually
 if [ "$ENV" == "PROD" ] && [ "${DISABLE_MIGRATION_SCRIPT_ON_PRODUCTION}" != "true" ]; then
-    scripts/migrate.sh
+    python3 manage.py migrate
+    python3 manage.py update_integrator_permissions
```
rbovard, line 13: Just to be sure, it's on purpose to run `update_integrator_permissions` only with `migrate`, and in all cases run `compilemessages` and `collectstatic`

rbovard (approved):

rbovard (commented):
~And what about the `set -e` which was in the `migrate.sh`?~ It's already in this script

entrypoint.sh
```diff
@@ -6,9 +6,17 @@ set -e
 # If not will, the container will fail and restart.
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
-# On PROD, we run migrations at startup unless explicitly disabled.
+# On PROD, we run migrations at startup unless explicitly disabled. If disabled,
+# this commands must be run manually
```
rbovard, line 10:
```suggestion
# If disabled, these commands must be executed manually.
```

entrypoint.sh
```diff
@@ -6,9 +6,17 @@ set -e
 # If not will, the container will fail and restart.
 python3 manage.py shell -c "import django; django.db.connection.ensure_connection();"
 
-# On PROD, we run migrations at startup unless explicitly disabled.
+# On PROD, we run migrations at startup unless explicitly disabled. If disabled,
```
rbovard, line 9:
```suggestion
# On PROD, we run migrations at startup unless explicitly disabled.
```

entrypoint.sh
rbovard: As discussed with @monodo, we run `update_integrator_permissions` and `compilemessages` every time, regardless of theenvironment.

rbovard:
@monodo please review and merge :)

PR #934, opened 2024-02-05
use uuid for field names to ensure uniqueness

monodo:


geocity/tests/factories.py
```diff
@@ -463,7 +464,7 @@ class FieldFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = forms_models.Field
 
-    name = factory.Faker("word")
+    name = factory.LazyAttribute(lambda a: str(uuid.uuid4())[0:6])
```
rbovard, line 467: Is 6 enough to garantee uniqueness?

rbovard (approved):

PR #940, opened 2024-02-14
YC-1176: Hide geotime number if unique

rbovard:
Fix https://jira.liip.ch/browse/YC-1176

### DIFF
--- a/geocity/apps/core/static/js/submission_submit.js
+++ b/geocity/apps/core/static/js/submission_submit.js
@@ -8,3 +8,9 @@ checkbox.addEventListener("click", () => {
         button.toggleAttribute("disabled");
     }
 });
+
+let formNumber = document.querySelector("[data-geo-time-role='formNumber']");
+
+if (formNumber.textContent == "1") {
+  formNumber.style.display = "none";
+}


geocity/apps/core/static/js/submission_submit.js
```diff
@@ -8,3 +8,9 @@ checkbox.addEventListener("click", () => {
         button.toggleAttribute("disabled");
     }
 });
+
+let formNumber = document.querySelector("[data-geo-time-role='formNumber']");
+
+if (formNumber.textContent == "1") {
```
AlexandreJunod, line 14: Count number of elements instead of just hiding the "1".
Put in a function and add event listener on the button used to add new geotimes

AlexandreJunod (changes requested):

geocity/apps/core/static/js/submission_submit.js
rbovard: @AlexandreJunod In fact I did the same logic as you did:
1. https://github.com/yverdon/geocity/blob/main/geocity/apps/core/static/js/submissions_submit.js#L3
2. https://github.com/yverdon/geocity/blob/main/geocity/apps/core/static/js/submissions_geo_time.js#L13

AlexandreJunod (approved):

PR #946, opened 2024-02-28
fix entrypoint

monodo:


entrypoint.sh
```diff
@@ -15,10 +15,15 @@ fi
 # On PROD, we always collect statics
 if [ "$ENV" == "PROD" ]; then
     python3 manage.py collectstatic --no-input
+    python3 manage.py update_integrator_permissions
+    python3 manage.py compilemessages -l fr
+fi
+
+if [ "$ENV" == "DEV" ]; then
```
rbovard, line 22: Can you do an "else" or it's not possible?

entrypoint.sh
```diff
@@ -15,10 +15,15 @@ fi
 # On PROD, we always collect statics
```
rbovard, line 15: Update comment according to the code

rbovard (changes requested):

AlexandreJunod (approved):

entrypoint.sh
AlexandreJunod: Autre fix possible :

if [ "$ENV"]; then
    python3 manage.py update_integrator_permissions
    python3 manage.py compilemessages -l fr
fi


Car le code ne fonctionne pas sans donnée en partant de 0 avec le code actuel.

entrypoint.sh
rbovard: Or even remove it, the code is clear...

rbovard (approved):

entrypoint.sh
```diff
@@ -17,8 +17,11 @@ if [ "$ENV" == "PROD" ]; then
     python3 manage.py collectstatic --no-input
 fi
 
-python3 manage.py update_integrator_permissions
-python3 manage.py compilemessages -l fr
+# Run thoses commands on any instance (PROD, DEV, DEMO...)
+if [ "$ENV" ]; then
```
rbovard, line 21: But whats the diffrence with now, it will failed if the migrations are not ran before...

For me this commit is wrong.

rbovard (changes requested):

PR #948, opened 2024-02-28
Remove unnecessary migrate.sh

rbovard:
Unnecessary since #946


rbovard:
But since all the commands in the entrypoint, they are executed on the `docker-compose up` no?

At least we can remove it from the doc in the "Setting up full Docker non persistent demo".

rbovard:
@monodo Any thoughts on this one?

PR #951, opened 2024-03-04
add django-ckeditor-5

AlexandreJunod:


geocity/settings.py
```diff
@@ -568,6 +568,7 @@
 STATIC_ROOT = "/static_root"
 
 PRIVATE_MEDIA_ROOT = "/private_documents"
+MEDIA_ROOT = os.path.join(BASE_DIR, "private_documents")
```
monodo, line 571: ⚠️ Ensure this does not disclose all private documents

PR #960, opened 2024-03-26
Fix archive permissions

AlexandreJunod:
Prevent a validator with pilot group for other entity, to archive requests where he isn't pilot

### DIFF
--- a/geocity/apps/submissions/views.py
+++ b/geocity/apps/submissions/views.py
@@ -963,7 +963,7 @@ class ArchivedSubmissionListView(SingleTableMixin, ListView):
                         return JsonResponse(
                             data={
                                 "error": True,
-                                "message": self.permission_error_message,
+                                "message": f"{self.permission_error_message} (N°{submission_id})",
                             },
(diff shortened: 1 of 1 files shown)


geocity/apps/submissions/views.py
```diff
@@ -954,6 +954,20 @@ def archive(self):
...
+                        self.request.user, submission
+                    ):
+                        return JsonResponse(
+                            data={
+                                "error": True,
+                                "message": self.permission_error_message,
+                            },
+                            status=403,
+                        )
+
```
monodo, line 970: Explain which submission could not be archived to the user ? If many submissions are selected for archiving ans only one is causing the problem just list which could not be archived ? It's a nice to have, not a must.

AlexandreJunod: @monodo is that good like that ? Should we use `id` instead of `n°`
image

Trying to not add another loop just to list all submissions the user can't archive, but we can at least tell the one throwing an error

PR #965, opened 2024-03-27
Feature/yc 1181

AlexandreJunod:


geocity/apps/accounts/users.py
```diff
@@ -52,67 +52,50 @@ def get_integrator_permissions():
...
-    anonymous_users = []
-    for user in qs:
-        if remove_anonymous and user.userprofile.is_anonymous:
-            anonymous_users.append(user.pk)
-        elif (
-            user.userprofile.is_anonymous
-            and user.userprofile.administrative_entity.pk
-            not in integrator_administrative_entities_list
-        ):
-            anonymous_users.append(user.pk)
```
AlexandreJunod, line 115: Only part deleted, everything works the same. With DRY approach. 

Reason to remove this : We want to remove anonymous_users anyway, there's no reason to show anonymous users.

AlexandreJunod: Actually it is easier to read putting the 2 codes side by side, as I changed the order of almost everything

monodo (approved):

PR #966, opened 2024-03-27
fix display of backoffice info for author and validators

monodo:


rbovard (changes requested):
Thanks it does the job, nice approach.

When testing, I noticed an existing bug. As validator, we can see a field even if it's only visible by author because of the test at line 51 `{% if field_value.field.is_visible_by_author or field_value.field.is_visible_by_validators and is_validator %}`

I think we shoud use the same logical (visible by author and is author). Can you please adapt it?

monodo:
Fixed

AlexandreJunod (approved):
Code is OK for me

AlexandreJunod:
In PO review @rbovard 

https://jira.liip.ch/browse/YC-1080

rbovard (approved):
LGTM, thanks!

PR #996, opened 2024-05-13
Fix Compose file: add no_cache option on static_builder service and update compose v1 to v2

nicolas-sitylb:


nicolas-sitylb:
@rbovard yes, Compose v1 (aka the `docker-compose` command*) is deprecated and does not receive any updates since July 2023. 

\* `docker-compose` (with a dash) may _perhaps_ reference Compose v2 as an alias, but it can be confusing as it is also the compose v1 historical command.

Therefore, I would strongly encourage people to [start using Compose v2](https://docs.docker.com/compose/migrate/),  aka `docker compose` without a dash (which is now also integrated by default in Docker Desktop) as soon as they can. 

But I noticed that the github CI is apparently based on an old Docker version because it fails and I don't know how to setup the github CI environment...
Ideally, I would recommend that CI or quality check environments stick to the same environment as the one used in production, for obvious reasons.

rbovard:
Okay, it was mainly to make the commit and the PR title reflect what had been done.

OdyX:
This is now https://gitlab.com/geocity/geocity/-/merge_requests/10/diffs

PR #1000, opened 2024-05-21
backend check for additional file extension restrictions

monodo:
Quick implementation of an additional backend check for cases where the browser do not recognize the additional extension restrictions.

TODOs @AlexandreJunod :

- [x] Bring the validation error correctly to the user


AlexandreJunod:
@monodo Canot reproduce the error in a test, as we don't know the real reason of the error, and it's probably due to a web navigator. The actual version of the code will output the following in case form validation isn't working.

`messages.error(
                    request,
                    _("Une erreur est survenue lors de l'upload de fichier."),
                )`

PR #1003, opened 2024-05-28
[YC-1205] fix call to url that does not exist

monodo:


geocity/apps/accounts/admin.py
```diff
@@ -1011,13 +1011,11 @@ def sortable_str(self, obj):
 
     @admin.display(boolean=True)
     def has_background_image(self, obj):
-        try:
-            return obj.background_image.url is not None
-        except ValueError:
-            return False
+        return False if obj.background_image.name == "" else True
```
AlexandreJunod, line 1014: return False if obj.background_image.name else True

AlexandreJunod (changes requested):

AlexandreJunod (approved):

PR #1004, opened 2024-05-28
fix error on ftsearch

monodo:


geocity/apps/submissions/admin.py
```diff
@@ -258,29 +262,39 @@ def get_list_display(self, request):
         return list_display
 
     # Fields used in search_fields and list_filter
-    integrator_fields = [
+    superadmin_search_fields = [
```
AlexandreJunod, line 265: Superuser not superadmin. Keep correct naming to prevent mistakes like "pilot" and "secrétariat"

geocity/apps/submissions/admin.py
```diff
@@ -258,29 +262,39 @@ def get_list_display(self, request):
...
+        "name",
+        "form__name",
+        "integrator__name",
+    ]
+    integrator_search_fields = [
+        "name",
+        "form__name",
+    ]
+
+    superadmin_list_search_fields = [
```
AlexandreJunod, line 275: Same

AlexandreJunod (changes requested):

geocity/apps/submissions/admin.py
rbovard: FMY: what's the difference between "pilot" and "secrétariat"?

monodo: There is no difference. We should standardize that, also in doc.

monodo:
Done

geocity/apps/submissions/admin.py
```diff
@@ -258,29 +262,39 @@ def get_list_display(self, request):
         return list_display
 
     # Fields used in search_fields and list_filter
-    integrator_fields = [
+    superusersearch_fields = [
```
AlexandreJunod, line 265: superuser_search_fields* and we are good to go

AlexandreJunod (changes requested):

PR #1022, opened 2024-06-21
differenciate comment visible in resume and report

AlexandreJunod:


OdyX:
This is now https://gitlab.com/geocity/geocity/-/merge_requests/7

